PR #48, opened 2021-01-11
Refactoring Commit

themantre:
- Using validator number instead of address inside commit
- Add block hash to commit
- New calculation for committers hash


themantre:
Related to: https://github.com/zarbchain/zips/blob/main/zip-001.md

PR #95, opened 2021-03-10
implemented GetNetworkInfo gRPC call

Ali-Meh:
update proto
update static
added getNetworkInfo
added add peer to syncMock


Ali-Meh:
looks like there is a race error in between 2 lines below
may need more investihation
[l159 private_key.go](https://github.com/zarbchain/zarb-go/blob/main/crypto/private_key.go#L159)
[l161 private_key.go](https://github.com/zarbchain/zarb-go/blob/main/crypto/private_key.go#L161)

Issue #116, opened 2021-03-24
New method gRPC: Validators


b00f:
Adding new method to list all the Validators in gRPC.

Issue #115, opened 2021-03-24
New method for gRPC: Validator by number


b00f:
Adding new method to get validator info by  number  in gRPC.

PR #120, opened 2021-03-28
gRPC validator RPCs Impelmented

Ali-Meh:
- Refactor Validator API Calls
- Added validator message to proto
- Updated tests
- Implemented issues #116 and #115


b00f (approved):
Thanks @Ali-Meh

PR #123, opened 2021-04-04
Unbond and withdrawal Transactions

Ali-Meh:
* Implemented Unbonding executor
* Implemented Unbonding payload
* Replaced `Stack()` with `Power()`

* * * 

- API
	- [x] Add Tx unbonding Support to CLI
	- [x] Add Tx unbonding Support to gRPC
	- [x] Add Tx unbonding Support to capap
	- [x] Add Tx unbonding Support to http
	- [x] make `www/grpc/transaction.go` errors verbose
	- [x] Add Tx withdraw Support to CLI
	- [x] Add Tx withdraw Support to gRPC
- Logic
	- [x] Unbond trx fee and if we need to charge the validator acc or unbonding account which may need another sig added to payload to make sure the unbonder has required keys
	    - the unbound trx is free of charge
	- [x] Should it fire out validator or just error out when validator is in committee
        - [prevent validator unbound trx from proccessing while in committee](https://github.com/zarbchain/zarb-go/pull/123#discussion_r615971421)
	- [x] After unbonded the withdraw trx should it require fee or not?
		- will act like a `send` transaction and will have `fee` calculated acordingly
	- [x] Distinction between `Stack()` and `Power()`
- Withdraw 
	- [x] Allow Validator to withdraw unbonded Stake


cmd/zarb/tx/send_tx.go
```diff
@@ -66,7 +66,7 @@ func SendTx() func(c *cli.Cmd) {
 			seq = *seqOpt
 
 			if *amountOpt == 0 {
-				cmd.PrintWarnMsg("Stake is not defined.")
+				cmd.PrintWarnMsg("Amount is not defined.")
```
b00f, line 69: Thanks.

execution/executor/unbond.go
```diff
@@ -0,0 +1,49 @@
...
+		return errors.Errorf(errors.ErrInvalidTx, "Validator is in committee right now please wait")
+	}
+	if val.Sequence()+1 != trx.Sequence() {
+		return errors.Errorf(errors.ErrInvalidTx, "Invalid sequence. Expected: %v, got: %v", val.Sequence()+1, trx.Sequence())
+	}
+	//should use the vlidator stake to pay fee??
+	//can make the unbonding free
+
+	val.IncSequence()
+	// unbondingVal.SubtractFromBalance(trx.Fee())
```
b00f, line 38: Look at above comment

state/state.go
```diff
@@ -146,7 +146,7 @@ func (st *state) tryLoadLastInfo() error {
 
 	totalStake := int64(0)
 	st.store.IterateValidators(func(val *validator.Validator) (stop bool) {
-		totalStake += val.Stake()
+		totalStake += val.Power()
```
b00f, line 149: Stake is the same as Power for all validators except for Bootstrap validators.

The network starts with 4 bootstrap validators which don't have any stake in advance. So their power sets to one. This helps consensus to start the bootstrapping phase.

Note that Zarb doesn't have any pre-minted or pre-allocated tokens.

tx/tx.go
```diff
@@ -32,15 +32,15 @@ type txData struct {
...
-func (tx *Tx) Version() int                     { return tx.data.Version }
-func (tx *Tx) Stamp() crypto.Hash               { return tx.data.Stamp }
-func (tx *Tx) Sequence() int                    { return tx.data.Sequence }
-func (tx *Tx) PayloadType() payload.PayloadType { return tx.data.Type }
-func (tx *Tx) Payload() payload.Payload         { return tx.data.Payload }
-func (tx *Tx) Fee() int64                       { return tx.data.Fee }
-func (tx *Tx) Memo() string                     { return tx.data.Memo }
-func (tx *Tx) PublicKey() *crypto.PublicKey     { return tx.data.PublicKey }
-func (tx *Tx) Signature() *crypto.Signature     { return tx.data.Signature }
+func (tx *Tx) Version() int                     { return tx.data.Version }   // nodes version generating transaction
```
b00f, line 35: Transaction version is not same as node version.

validator/validator.go
```diff
@@ -39,12 +40,15 @@ func (val *Validator) Number() int                 { return val.data.Number }
...
 func (val *Validator) Stake() int64                { return val.data.Stake }
 func (val *Validator) BondingHeight() int          { return val.data.BondingHeight }
+func (val *Validator) UnbondingHeight() int        { return val.data.UnbondingHeight }
 func (val *Validator) LastJoinedHeight() int       { return val.data.LastJoinedHeight }
 
 func (val Validator) Power() int64 {
 	// Only bootstrap validators at genesis block has no stake
 	if val.data.Stake == 0 {
 		return 1
+	} else if val.data.UnbondingHeight > 0 { //if the validator requested to unbound ignore stake
```
b00f, line 50: We don't need to unbond bootstrap validators because they don't have any stake. So this check is redundant.

tx/factory.go
```diff
@@ -60,6 +60,25 @@ func NewBondTx(stamp crypto.Hash,
...
+	return &Tx{
+		data: txData{
+			Stamp:    stamp,
+			Sequence: seq,
+			Version:  1,
+			Type:     payload.PayloadTypeUnbond,
+			Payload: &payload.UnbondPayload{
+				Validator: val,
+			},
+			Fee:  0,
```
b00f, line 76: Good. Let's keep it zero. Unbond transactions are fee free.

b00f (changes requested):

execution/executor/sortition.go
```diff
@@ -22,6 +22,9 @@ func (e *SortitionExecutor) Execute(trx *tx.Tx, sb sandbox.Sandbox) error {
 	if val == nil {
 		return errors.Errorf(errors.ErrInvalidTx, "Unable to retrieve validator")
 	}
```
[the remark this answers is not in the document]
b00f: In some blockchain validators can rebond, but this is not safe. Imagine a hacker has access to the validator key, then it can rebond and send malicious votes to the network.

state/state.go
Ali-Meh: > Stake is the same as Power for all validators except for Bootstrap validators.
exactly for this they should be used cautiously and the they should be distinguished as result of this exception i passed a [test](https://github.com/zarbchain/zarb-go/pull/123/files/6eac6392c331973bd133346ed78df485e34e0537#diff-d16667d244a9c4e4211ee05f2d187488343c959396a63891f179ce5eaf438e04L64) `around` witch may raise issues along the way ,unlikely but possible

execution/executor/sortition.go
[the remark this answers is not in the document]
Ali-Meh: [the lines below ](https://github.com/zarbchain/zarb-go/pull/123/files/6eac6392c331973bd133346ed78df485e34e0537#diff-4052820177a99798ca41230c6ab7c97f271a525379b0a2d831258c0c05a822f7R25-R27) check if power is 0 i wanted to have distinction of stack and power for this reason actually the power when ever `UnbondingHeight ` is set will return 0 indicating the fact validator has unbounded

```go
func (val Validator) Power() int64 {
	// Only bootstrap validators at genesis block has no stake
	if val.data.Stake == 0 {
		return 1
	} else if val.data.UnbondingHeight > 0 { //if the validator requested to unbound ignore stake
		return 0
	}
	return val.data.Stake
}
```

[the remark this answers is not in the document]
b00f: Yes. I got it. This is nice. By this solution, you actually set validator power to zero. It looks its car engine stopped. Keep it. 

Only add this logic to Bond execution.

state/state.go
b00f: TotalStake used for VRF and sortition. What is your error?

b00f (changes requested):

state/state.go
Ali-Meh: > 
> 
> TotalStake used for VRF and sortition. What is your error?

it might be just test env but may happen in slow networks as well that the newly generated validators always have stake 1!

b00f: > it might be just test env but may happen in slow networks as well that the newly generated validators always have stake 1

Stake for new validators is Zero. Power only applies inside consensus.

execution/executor/sortition.go
```diff
@@ -22,6 +22,9 @@ func (e *SortitionExecutor) Execute(trx *tx.Tx, sb sandbox.Sandbox) error {
 	if val == nil {
 		return errors.Errorf(errors.ErrInvalidTx, "unable to retrieve validator")
 	}
+	if val.Power() == 0 {
```
b00f, line 25: Please add this comment above this statement: "Power for parked validators are set to zero"

execution/executor/unbond.go
```diff
@@ -0,0 +1,47 @@
...
+		return errors.Errorf(errors.ErrInvalidTx, "Unable to retrieve validator assoiciated with this key")
+	}
+	if e.strict && sb.IsInCommittee(pld.Validator) {
+		return errors.Errorf(errors.ErrInvalidTx, "Validator is in committee right now please wait")
+	}
+	if val.Sequence()+1 != trx.Sequence() {
+		return errors.Errorf(errors.ErrInvalidTx, "Invalid sequence. Expected: %v, got: %v", val.Sequence()+1, trx.Sequence())
+	}
+	if val.UnbondingHeight() > 0 {
+		return errors.Errorf(errors.ErrInvalidTx, "you already have unbonded at Height %v", val.UnbondingHeight())
```
b00f, line 34: Can you cover this line?

tx/tx.go
```diff
@@ -33,15 +33,15 @@ type txData struct {
...
-func (tx *Tx) Stamp() crypto.Hash           { return tx.data.Stamp }
-func (tx *Tx) Sequence() int                { return tx.data.Sequence }
-func (tx *Tx) PayloadType() payload.Type    { return tx.data.Type }
-func (tx *Tx) Payload() payload.Payload     { return tx.data.Payload }
-func (tx *Tx) Fee() int64                   { return tx.data.Fee }
-func (tx *Tx) Memo() string                 { return tx.data.Memo }
-func (tx *Tx) PublicKey() *crypto.PublicKey { return tx.data.PublicKey }
-func (tx *Tx) Signature() *crypto.Signature { return tx.data.Signature }
+func (tx *Tx) Version() int                 { return tx.data.Version }   // Transaction version
+func (tx *Tx) Stamp() crypto.Hash           { return tx.data.Stamp }     // hash of previus block
```
b00f, line 37: I think these comments are not necessary. Code itself should talk. 
Here  some comments are not accurate like stamp or sequence. Let's keep the code uncommented.

tx/tx_test.go
```diff
@@ -296,6 +296,39 @@ func TestBondSignBytes(t *testing.T) {
 	assert.True(t, trx1.IsBondTx())
 }
 
+func TestUnbondSignBytes(t *testing.T) {
+	h := crypto.GenerateTestHash()
+	signer := crypto.GenerateTestSigner()
+	// _, pub, _ := crypto.GenerateTestKeyPair()
```
b00f, line 302: Please remove this line.

b00f (commented):
LGTM. We can merge this PR now. 
Thanks

b00f (approved):

b00f:
@ali-mahdavi99 Thanks.

PR #127, opened 2021-04-07
gRPC APIs Implmented

Ali-Meh:
- Recompile statik
- Add address to validator reponse 
- GetBlockHieght gRPC API Implemented
- GetBlockchainInfo gRPC API Implemented
- SendRawTransaction gRPC  API Implemented


state/mock.go
```diff
@@ -26,14 +27,18 @@ type MockState struct {
...
-		GenHash:   crypto.GenerateTestHash(),
-		Store:     store.MockingStore(),
-		TxPool:    txpool.MockingTxPool(),
-		Committee: committee,
+		GenHash:     crypto.GenerateTestHash(),
+		Store:       store.MockingStore(),
+		TxPool:      txpool.MockingTxPool(),
+		Committee:   committee,
+		AddToPool:   true,
+		BroadCastTx: true,
```
b00f, line 41: About `AddToPool` and `BroadCastTx`:

I have a suggestion here. If a transaction already exists inside pool, Add and Broadcast should return an error.
By this check we can omit these two variables which look unnecessary to me.

b00f (commented):
LGTM.

The only comment I have here is new boolean variables inside txPool.Mock struct.

PR #130, opened 2021-04-09
State db

yenkhoon:
- using dependency injection (passing db instance) design pattern
- now we support only one store for all


b00f:
@yenkhoon There are some linting issues here. To fix formatting and linting issues please run this command:
```
make fmt
```

b00f (approved):

PR #144, opened 2021-04-25
Added CORS to config

nagarajmanjunath:
* Added the CORS to grpc gateway.
* Added test case for grpc cors enable


www/grpc/server_test.go
```diff
@@ -19,7 +19,6 @@ var tMockState *state.MockState
 var tMockSync *sync.MockSync
 var tListener *bufconn.Listener
 var tCtx context.Context
-var tGrpc Config
```
b00f: Why did you add gRPC here?

b00f (changes requested):

www/grpc/server_test.go
nagarajmanjunath: i have removed it.

PR #159, opened 2021-05-01
p2p: Drop connection for banned peers

themantre:
- Add new field to peer to hold the status of the peer
- If a peer is banned, drop the connection
- Disconnect function added to network module


d4rk5h0t (approved):
awsome lgtm

PR #165, opened 2021-05-04
automate binary releases

d4rk5h0t:
- update version module
    -  ~~added SemVer to version~~
    - ~~auto generate version from tag assignment via `git tag v1.0.0-beta` or `git tag v1.0.0`~~
- add goreleaser
    -  creating tag will result in genrating release artifacts for win,linux,macos/amd64
    - push tags via `git push --tag origin` and the "v*" format customizable in `.github\workflows\release.yml`
- removed bls compilation requirements~~
    - ~~now using prebuilt [bls-go-binary](https://github.com/herumi/bls-go-binary/commit/19c7242910d2e1696d6a1c85440c26d9024cacf8)~~
    - ~~it might be better idea to abandon compiling bls repo and use the commit above as nested repo or maybe use it only for releases and keep compiling for docker and development~~


.github/workflows/bls.yml
```diff
@@ -1,17 +0,0 @@
-on: [pull_request]
```
b00f, line 1: Why this workflow is deleted. We want to make sure Zarb can be compiled with bls library.

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
+name: releaser
+
+on:
+  create:
+      tags:
+  # push:
```
b00f, line 6: Please remove these comments.

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+      -
+        name: Cache Go modules
+        uses: actions/cache@v1
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+            
+      # - name: Setup tmate session
```
b00f, line 57: Better put a comment that why these lines are commented or remove these commented lines. ;)

Makefile
```diff
@@ -22,24 +32,23 @@ tools:
 	go mod tidy
 
 bls:
-	@echo "Compiling bls"
+	@echo "cloning bls"
 	rm -rf $(HERUMI)
-	git clone --recursive git://github.com/herumi/bls.git $(HERUMI)/bls && cd $(HERUMI)/bls && make minimized_static
+	git clone https://github.com/herumi/bls-go-binary.git $(HERUMI)
```
b00f, line 37: MAC users with M1 architecture need to compile BLS manually.

In case we need to use compiled version, we don't need this target.

Makefile
```diff
@@ -1,10 +1,20 @@
 PACKAGES=$(shell go list ./... | grep -v 'tests')
-HERUMI= $(shell pwd)/.herumi
-BLS_CGO_LDFLAGS=CGO_LDFLAGS="-L$(HERUMI)/bls/lib -lbls384_256 -lm -lstdc++"
-BUILD_LDFLAGS= -ldflags "-X github.com/zarbchain/zarb-go/version.build=`git rev-parse --short=8 HEAD`"
 
+TAGS="zarb"
+VTAG=$(shell git describe --tags --abbrev=0)
+HERUMI= .herumi
+LDFLAGS= -ldflags "-X github.com/zarbchain/zarb-go/version.build=`git rev-parse --short=8 HEAD` -X github.com/zarbchain/zarb-go/version.semVer=$(VTAG)"
```
b00f, line 6: This is really confusing. See my comments about SemVer package down.

Makefile
```diff
@@ -1,10 +1,20 @@
...
 
 
+_OS=$(shell go env GOOS)
+_ARCH=$(shell go env GOARCH)
+
+
+LIB_DIR=$(shell pwd)/$(HERUMI)/bls/lib/$(_OS)/$(_ARCH)
+CGO_LDFLAGS=CGO_LDFLAGS="-L$(LIB_DIR) -lbls384_256 -lm -lstdc++"
+CGO_LDFLAGS_Default=$(shell go env CGO_CFLAGS)
```
b00f, line 15: `CGO_LDFLAGS_Default=$(shell go env CGO_LDFLAGS)` is better.
`CFLags` mostly uses for compiling and `LDFLAGS` for linking. A bit different.

Makefile
```diff
@@ -22,24 +32,23 @@ tools:
...
+	go build $(LDFLAGS) -tags $(TAGS) -o build/zarb ./cmd/zarb/
 
 install:
-	go install $(BUILD_LDFLAGS) ./cmd/zarb
-
-build_bls:
-	$(BLS_CGO_LDFLAGS) go build $(BUILD_LDFLAGS) -o build/zarb ./cmd/zarb/
+	go install $(LDFLAGS) -tags $(TAGS) ./cmd/zarb
 
-build_bls_release:
```
b00f, line 41: This target is used inside docker.

Makefile
```diff
@@ -1,10 +1,20 @@
...
+HERUMI= .herumi
+LDFLAGS= -ldflags "-X github.com/zarbchain/zarb-go/version.build=`git rev-parse --short=8 HEAD` -X github.com/zarbchain/zarb-go/version.semVer=$(VTAG)"
 
 
+_OS=$(shell go env GOOS)
+_ARCH=$(shell go env GOARCH)
+
+
+LIB_DIR=$(shell pwd)/$(HERUMI)/bls/lib/$(_OS)/$(_ARCH)
+CGO_LDFLAGS=CGO_LDFLAGS="-L$(LIB_DIR) -lbls384_256 -lm -lstdc++"
```
b00f, line 14: Not bad we add optimization flags here: 
`CGO_LDFLAGS=CGO_LDFLAGS="-L$(LIB_DIR) -lbls384_256 -lm -lstdc++ -g -O2"`

Makefile
```diff
@@ -1,10 +1,20 @@
...
+VTAG=$(shell git describe --tags --abbrev=0)
+HERUMI= .herumi
+LDFLAGS= -ldflags "-X github.com/zarbchain/zarb-go/version.build=`git rev-parse --short=8 HEAD` -X github.com/zarbchain/zarb-go/version.semVer=$(VTAG)"
 
 
+_OS=$(shell go env GOOS)
+_ARCH=$(shell go env GOARCH)
+
+
+LIB_DIR=$(shell pwd)/$(HERUMI)/bls/lib/$(_OS)/$(_ARCH)
```
b00f, line 13: I believe compiling BLS is more safe than using Herumi's compiled version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+jobs:
+  build:
+    name: Build release on ${{ matrix.os }} with ${{ matrix.compiler }}.
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
b00f, line 21: I prefer giving version. Like ubuntu-18.

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+            
+      # - name: Setup tmate session
+      #   uses: mxschmitt/action-tmate@v3
+
+      - name: Get the version
+        id: get_version
+        run: |
```
b00f, line 62: get version by `go run cmd/zarb Version`

b00f:
@d4rk5h0t Thanks for this amazing PR. Here is my toughts:

- Try to get version by running Zarb. You may build Zarb and run `zarb version` or use `go run ...`
- We don't compile BLS in most of the time. For some platform compiling BLS is the only way to build Zarb.

version/version.go
```diff
@@ -2,17 +2,31 @@ package version
...
 	minor uint   = 0
 	patch uint   = 0
 	meta  string = "beta"
 )
 
 var build string
+var semVer string
+
+func init() {
+	sv, err := semver.NewVersion(semVer)
```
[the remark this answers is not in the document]
d4rk5h0t: i didn't get your point if we gonna have multiple LTS version then i would suggest having distinct branch for every minor version at least like other like [ethereum](https://github.com/ethereum/go-ethereum/tree/release/1.0.1) does.

besides I'm just using semver package to safely parse the build time tag set on package so there be no need to manually change source for every release and via `git tag ` command and the ` -X github.com/zarbchain/zarb-go/version.semVer=$(VTAG)` ldflag it will be set automatically

.github/workflows/bls.yml
d4rk5h0t: > it might be better idea to abandon compiling bls repo and use the commit above as nested repo or maybe use it only for releases and keep compiling for docker and development

as i suggested in the pull request  comment i deleted it to show case abandoning it

Makefile
d4rk5h0t: then i guess better to separate release pipe from compiling pipe

Makefile
d4rk5h0t: i might need to add comments to clarify it
but it utilizes git tag command to dynamically set build version to latest tag has been set

Makefile
d4rk5h0t: ohh shoot🤦‍♂️
absolutely right

Makefile
d4rk5h0t: as for release pipeline its more convenient and faster but if we separate pipelines there should be no issue

.github/workflows/release.yml
d4rk5h0t: that is not feasible at this step it's not been set so it default to `v1.0.0-beta` 
the version is set in next step via `goreleaser` action so actually we don't need this any more

d4rk5h0t:
> 
> 
> @d4rk5h0t Thanks for this amazing PR. Here is my toughts:
> 
>     * Try to get version by running Zarb. You may build Zarb and run `zarb version` or use `go run ...`
> 
>     * We don't compile BLS in most of the time. For some platforms compiling BLS is the only way to build Zarb.

- on first issue well I'm trying to automate versioning of package so i think thats out of the window to use zarb hard coded version
- i would suggest to separate a pipeline for releasing on github with use of `git tag` and pre-built bls archives

version/version.go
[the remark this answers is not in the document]
themantre: @d4rk5h0t

Yes, As you said we will have a separate and distinct branch per release (or tag).

In your code, Version depends on someone else (`git describe` for example). Right? Think crazy, maybe git doesn't work fine in this case. So we have trouble then.

My point is, when version defines manually, we have full control over that, even we can break the rules!

themantre:
@d4rk5h0t 

Let me know your idea about versioning tag. We can discuss about Pipeline and BLS precompiled binaries after that.

.gitignore
```diff
@@ -7,5 +7,8 @@ vendor
 .vscode
 todo
 local.mk
+*.local.md
 .herumi
+dist/
+.releaser.yml
```
b00f, line 13: Do we need this line?

Ali-Meh: yeap actually the build files will reside in dist and `.releaser.yml` will be auto generated for os specific command

version/version.go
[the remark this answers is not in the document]
d4rk5h0t: you absolutely right 
but let say git fails or the tag is not following semver specification 
then as shown in code the `init()` function only runs only if tag setted on commit is parsable by semver package (semver standard) otherwise it will default to `v1.0.0-beta` as hardcoded in code witch I'll strongly recommend to set for every release branch

.gitignore
d4rk5h0t: yes actully the binaries generated will reside in `dist` and `.releaser.yml` will be generated for OS specific env

d4rk5h0t:
> 
> 
> @d4rk5h0t
> 
> ~Let me know your idea about versioning tag. We can discuss about Pipeline and BLS precompiled binaries after that.~
> 
> We are going to release version 1.0.0 soon and it's better we avoid some unnecessary changes right now. Please revert `makefile` & `version.go`. If releaser can't work for MAC machines, we can ignore it now. We will fix it for the next version.
> 
> Thanks

no problem it was just a convenient way to get binaries out as fast as possible and I actually separated command used for compiling and releasing but feel free to close it👍

version/version.go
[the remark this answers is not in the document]
b00f: This is a bit confusing. Git tag is not designed for versioning control and therefore better to not use them as a source of versioning.

.goreleaser.yml
```diff
@@ -0,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+project_name: zarb-cli
```
b00f, line 3: Right now Zarb binary is  daemon and CLI both. In the next version we might have different binaries for that. Since then could you please rename `zadb-cli` to `zarb`.

.goreleaser.yml
```diff
@@ -0,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sane defaults.
```
b00f, line 1: What is your idea moving this file into .github folder? next to `release.yml`?

.goreleaser.yml
d4rk5h0t: sure

.goreleaser.yml
d4rk5h0t: won't suggest it since by editing `${GOHOSTOS}` or using the command 
```bash
GOHOSTOS=$(go env GOHOSTOS) envsubst < ".goreleaser.yml" > .releaser.yml && release --rm-dist -f .releaser.yml
```
the can build it on their local machine besides its not github action config rather goreleaser app config file

b00f (approved):

PR #167, opened 2021-05-04
p2p: Add maximum session parameter to config

b00f:
- Adding maximum session number to config
- Check if the node is busy before parsing the block requests
- Check if open sessions are not more than maximum number


b00f:
```
INFO[0003] New block is committed                        _state="{#252855 ⌘ 84E863AC19AC 🕣 22.08.40}" block="{⌘ 84E863AC19AC 👤 zrb1jwqeg36r 💻 28A0DD3C4A2E 📨 1}" round=0
INFO[0003] New block is committed                        _state="{#252856 ⌘ F84964635A16 🕣 22.08.50}" block="{⌘ F84964635A16 👤 zrb1eycmsfal 💻 604FC6698EEB 📨 1}" round=0
INFO[0003] New block is committed                        _state="{#252857 ⌘ FA980CBF3656 🕣 22.09.10}" block="{⌘ FA980CBF3656 👤 zrb1gdke55ha 💻 9BB1CAD7C25E 📨 1}" round=1
INFO[0003] New block is committed                        _state="{#252858 ⌘ 8C085BC411C7 🕣 22.09.20}" block="{⌘ 8C085BC411C7 👤 zrb1drqa3zpy 💻 4879FD38DDDA 📨 1}" round=0
INFO[0003] New block is committed                        _state="{#252859 ⌘ EB23D97A6161 🕣 22.09.30}" block="{⌘ EB23D97A6161 👤 zrb1ffkakjps 💻 1658D531E57F 📨 1}" round=0
```

This PR will mitigate time jump (10 seconds) if a validator is offline or faulty. Look at the above blocks time.

sync/handler_download_request.go
```diff
@@ -52,6 +77,8 @@ func (handler *downloadRequestHandler) ParsPayload(p payload.Payload, initiator
 			break
 		}
 	}
+	// To avoid sending blocks again, we update height for this peer
+	peer.UpdateHeight(from - 1)
```
d4rk5h0t, line 81: given its decentralized network i don't see how sure we are about applicant getting the response

sync/sync.go
```diff
@@ -223,8 +223,9 @@ func (sync *synchronizer) Fingerprint() string {
...
 }
 
-func (sync *synchronizer) sendBlocksRequestIfWeAreBehind() {
-	if sync.peerSet.HasAnyValidSession() {
+func (sync *synchronizer) checkIfWeAreBehindTheNetwork() {
+	// TODO: write test for me
+	if sync.peerSet.HasAnyOpenSession() {
 		sync.logger.Debug("We have open seasson")
 		return
 	}
```
d4rk5h0t, line 231: I would suggest renaming this func to more descriptive one like `UpdateNodeHeight` or `UpdateNodeChain` since it actually requests blocks and joins topics

sync/sync.go
```diff
@@ -241,11 +242,11 @@ func (sync *synchronizer) sendBlocksRequestIfWeAreBehind() {
 			if err := sync.network.JoinDownloadTopic(); err != nil {
 				sync.logger.Info("We can't join download topic", "err", err)
 			} else {
-				sync.RequestForMoreBlock()
+				sync.downloadBlocks()
 			}
 		} else {
 			sync.logger.Info("We are behind the network, Ask for more blocks")
```
d4rk5h0t, line 248: `behind the network` feel like supporting network I'd suggest something like `less than network height ` or `shorter than network`

sync/peerset/peer_set.go
```diff
@@ -59,18 +60,31 @@ func (ps *PeerSet) FindSession(id int) *Session {
...
+func (ps *PeerSet) HasAnyOpenSession() bool {
 	ps.lk.Lock()
 	defer ps.lk.Unlock()
 
+	ps.removeExpiredSessions()
+
+	return len(ps.sessions) != 0
+}
+
+func (ps *PeerSet) removeExpiredSessions() {
```
d4rk5h0t, line 81: since both `HasAnyOpenSession` and `NumberOfOpenSessions`  are tempering with `ps.sessions` i don't see why there lucks are different i think both need Read and Write protective lock

d4rk5h0t (commented):
awesome pull just some thought to improve it

- might be a better to substitute `behind the network` with something like `shorter than network` or `less than network height` for more clarification.
-  since the applicant of blocks has no idea of the nodes `RequestBlockInterval` i would suggest to publish it on network beforehand or default it to something
- fixed number of sessions might cause in closed network topology it might be good idea to randomize it so we might get some bridges

sync/peerset/peer_set.go
b00f: Thanks, Fixed.

b00f:
> fixed number of sessions might cause in closed network topology it might be good idea to randomize it so we might get some bridges

Number of open sessions should be customizable based on CPU and memory of the node. If a node has more computational power and memory capacity, it can respond to more requests (or sessions).

b00f:
> since the applicant of blocks has no idea of the nodes RequestBlockInterval i would suggest to publish it on network beforehand or default it to something

Good point. This parameter now is default and set to 720 blocks (about two hours)

sync/sync.go
b00f: `updateBlockchain` make more sense. I update the code

sync/handler_download_request.go
```diff
@@ -21,13 +22,37 @@ func (handler *downloadRequestHandler) ParsPayload(p payload.Payload, initiator
...
+	}
+
+	if peer.Height() > pld.From {
+		response := payload.NewDownloadResponsePayload(payload.ResponseCodeRejected, pld.SessionID, initiator, 0, nil, nil)
+		handler.broadcast(response)
+
+		return errors.Errorf(errors.ErrInvalidMessage, "Peer request for blocks that already has: %v", pld.From)
+	}
+
 	if pld.To-pld.From > handler.config.RequestBlockInterval {
```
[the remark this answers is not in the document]
b00f: Now it's default and set to 720 (about two hours)

sync/sync.go
b00f: Thanks. Done.

sync/handler_download_request.go
b00f: We need to have a mechanism to prevent sending blocks to a node repeatedly. In a decentralize network this make sense, A node can ask other nodes to send the same blocks but not from the one who asked before.

sync/handler_latest_block_request.go
```diff
@@ -46,6 +71,8 @@ func (handler *latestBlocksRequestHandler) ParsPayload(p payload.Payload, initia
 
 		from += len(blocks)
 	}
+	// To avoid sending blocks again, we update height for this peer
```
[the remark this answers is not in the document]
b00f: We can't be sure. Sync module is based on Pulling mechanism. The node should always pull the blocks if it needs.

sync/handler_download_request.go
d4rk5h0t: @b00f 
sure but as the network is too small to have many archive nodes i think we need to make exceptions

consensus/height_test.go
```diff
@@ -63,7 +63,7 @@ func TestUpdateCertificate(t *testing.T) {
 	assert.Empty(t, cert.Absences())
 }
 
-func TestConsensusBehindState(t *testing.T) {
+func TestConsensusHeightIsShorterThanState(t *testing.T) {
```
d4rk5h0t, line 66: for future references might be good idea to keep pull request in titles scope and open separate PRs for typos like this (just a thought which i don't mind myself😉)

d4rk5h0t (approved):
ALL good just a thought that if out network is still premature to have enough Archive nodes supporting network and the possibility of message loss we might need to make some exceptions while the network is still small besides I'm not really sure how Gossipsub can make sure all messages make to the recipient

themantre (approved):

PR #181, opened 2021-05-09
Update Makefile and support Windows

b00f:
In this PR we are going to support windows OS. This PR includes these changes:

Generating random path is modified to fix a potential issue on windows. 
Also generating random number is more stronger (using `crypto/rand` instead of `math/rand`). It can prevent duplicated random numbers generation specially in Windows.
In order to simplifying Makefile, include path removed from capnp target.

**About `herumi` target in Makefile:**
[bls-go-binary](https://github.com/herumi/bls-go-binary) package has [precompiled libraries](https://github.com/herumi/bls-go-binary/tree/master/bls) for Windows, Linux and MaxOS. However for other OS/Arc like ARM or BSD users have to compile BLS library. We have two choices:
1- Detecting OS/ARC inside Makefile and decide what to do 
2- Building BLS before compiling Zarb for all OS/Arc

The first solution brings some complexity to the Makefile. I opted for the second choice but I am open to discuss on this. 

Hope this helps. Thanks


.gitignore
```diff
@@ -1,14 +1,12 @@
...
-.vscode/launch.json
 vendor
 .DS_Store
-.bls
 .vscode
 todo
-local.mk
-*.local.md
+local*
+.local*
```
d4rk5h0t, line 8: `*.local.*` might be better

Makefile
```diff
@@ -1,60 +1,53 @@
...
+	go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.3.0
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.25.0
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0
+	go install github.com/bufbuild/buf/cmd/buf@v0.39.1
+	go install github.com/rakyll/statik@v0.1.7
+
+
+herumi:
+	@if [ ! -d $(HERUMI) ]; then \
+		rm -rf $(HERUMI); \
```
d4rk5h0t, line 25: removing non existing path? why?

Makefile
```diff
@@ -1,60 +1,53 @@
...
-	@echo "Installing tools"
-	GO111MODULE=off go get zombiezen.com/go/capnproto2/...
-	go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.38.0
-	go get github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@v2.3.0
-	go get github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.3.0
-	go get google.golang.org/protobuf/cmd/protoc-gen-go@v1.25.0
-	go get google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0
-	go get github.com/bufbuild/buf/cmd/buf@v0.39.1
-	go get github.com/rakyll/statik@v0.1.7
-	go mod tidy
```
d4rk5h0t, line 23: why this line is deleted it's not a bad idea to clean up after we done

store/mock.go
```diff
@@ -104,7 +104,8 @@ func (m *MockStore) HasAnyBlock() bool {
 
 func (m *MockStore) IterateAccounts(consumer func(*account.Account) (stop bool)) {
 	for _, a := range m.Accounts {
-		stopped := consumer(&a)
+		acc := a
+		stopped := consumer(&acc)
 		if stopped {
 			return
 		}
```
d4rk5h0t, line 111: i don't see why do we need a extra redundant variable in thread safe functionality

at line 117 as well

.github/workflows/testing.yml
```diff
@@ -15,10 +19,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
-    - name: Build
-      run: make build
-
```
d4rk5h0t, line 20: since deleted `bls.yml` might be good to test it compileable as well on other OSs so avoid facing compile issues on external lib

d4rk5h0t (commented):
all good kudos on windows issues
just some changes might be helpfull

.gitignore
b00f: Done. Thanks

Makefile
b00f: Done. Thanks

store/mock.go
b00f: It's security linting error. We pass the object with its pointer to outside and it might be changed.

.github/workflows/testing.yml
b00f: All targets have BLS now.

themantre (approved):

PR #182, opened 2021-05-09
Update goreleaser

b00f:
- Remove snapshot  from goreleaser (we don't support snapshot)
- Formatting yml files
- Remove unrelated comments

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,25 +1,25 @@
-# This is an example .goreleaser.yml file with some sane defaults.
-# Make sure to check the documentation at http://goreleaser.com
 project_name: zarb
 
 before:
   hooks:
     - go mod tidy
+
 builds:
   - main: ./cmd/zarb/
     env:
       - CGO_ENABLED=1
-      - pwd = $(pwd)
-    # Custom ldflags templates.
-    # Default is `-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`.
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -1,25 +1,25 @@
...
       - CGO_ENABLED=1
-      - pwd = $(pwd)
-    # Custom ldflags templates.
-    # Default is `-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`.
-    # could use `-X github.com/zarbchain/zarb-go/version.semVer={{.Version}} -extldflags "-static"` to dynamicly version and staticly build(has some issues with mac)
+
+    # Omit the symbol table and debug information.
+    # For more info refer to: https://golang.org/cmd/link/
     ldflags:
-     - -s -w -X github.com/zarbchain/zarb-go/version.build={{.ShortCommit}}
```
d4rk5h0t, line 17: this could help us get to root  cause of an issue faster on github and help users make sure of sha hashes easier

d4rk5h0t (approved):
All good just a thought on version.build might be helpfull

.goreleaser.yml
b00f: For beta versions we have [build](https://github.com/zarbchain/zarb-go/blob/1c7fe7da18d70c9b59ab5e21b478bf2b89df96f1/version/version.go#L15) tag that is the short hash for the git commit. 
Released binaries are created from an existing tags and since tags are unique, we can find out which code the node is running.

By the way, I am going to close this PR because it's not helpful. I will open a discussion about releaser to discuss more about it.

Issue #185, opened 2021-05-10
How to send and recieve transaction?


khaninejad:
Hi,

I need a method to implement a send and receive in the Zarb blockchain network using other clients rather than the core, but I couldn't find it on documents.
I would be happy if you assist me.

Thank you

d4rk5h0t:
@khaninejad Thanks for bringing this up.

we do have gRPC, capnp, http(capnp) rpc endpoints to request a transaction or retrieve one
at the end of day you need a node running network to request these for you or a more unlikely run a node on phone itself and publish your transaction that way.
our `zarb` executable currently just generates the payload and lets you sign it we plan to add `server` flag to it so can publish it as well in the future.

our endpoints for sending or getting a transaction info are listed below i think you need to `generate payload -> serialize it -> sign it -> send it to a node` so there is not much we can do about it

```bash
rpc SendRawTransaction(SendRawTransactionRequest) returns(SendRawTransactionResponse)     { option (google.api.http).put = "/api/send_raw_transaction/{data}";};   #grpc endpoint for sending transaction()
rpc GetTransaction(TransactionRequest) returns (TransactionResponse)                      { option (google.api.http).get = "/api/tranaction/id/{id}";}            #grpc endpoint for retrieving one()
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
/send_raw_transaction/{data}         #http endpoint for sending transaction
/transaction/id/{hash}                     #http endpoint for retrieving one
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
sendRawTransaction   @7 (rawTx: Data)                            -> (result :SendTransactionResult); #capnp endpoint for sending transaction
getTransaction       @2 (id: Data, verbosity: Int32)             -> (result :TransactionResult);         #capnp endpoint for retrieving one
```

but to generate payloads you need `.proto` file and [Data serialization](https://zarb.network/guide/learn-serialization.html) of [payload](https://github.com/zarbchain/zarb-go/blob/main/tx/tx.go#L24)  and [send transaction payload](https://github.com/zarbchain/zarb-go/blob/main/tx/payload/send.go#L10) we have a working [sample](https://github.com/zarbchain/zarb-go/blob/main/cmd/zarb/tx/send_tx.go#L115-L117) in go for it

let me know if you still need more clarification.
@b00f feel free to correct any part you see fit

khaninejad:
[...]

PR #204, opened 2021-05-14
cmd: tx send command sends transaction to network

Ali-Meh:
## Description
seprate `tx send` from `payload send` in cmd for better UX

- Seprate sendtx and payload sendtx in cmd
- add  stamp to blockchian info stamp and update proto
- confirm to publish tx
- gRPC transaction function error made verbose
- Seprate payload and tx in cli
- Add gRPC Client in cmd
- Confirm to publish bondtx

## Related issue(s)
#185
#195

## Checklist

- [x] I have added tests to cover my changes.
- [x] CHANGELOG is updated.


cmd/gRPC.go
```diff
@@ -0,0 +1,63 @@
+package cmd
```
b00f, line 1: For the name of file is better to keep it all lowercase: "grpc.go"

cmd/zarb/main.go
```diff
@@ -21,9 +22,13 @@ func zarb() *cli.Cli {
...
 		k.Command("change-auth", "Change the passphrase of a keyfile", key.ChangeAuth())
 	})
-	app.Command("tx", "Create raw transaction", func(k *cli.Cmd) {
-		k.Command("bond", "Generate raw bond transaction", tx.BondTx())
-		k.Command("send", "Generate raw send transaction", tx.SendTx())
+	app.Command("tx", "Create, Sing, Publish transactions", func(k *cli.Cmd) {
+		k.Command("bond", "Send signed Bond transaction to network", tx.BondTx())
+		k.Command("send", "Send signed Send transaction to network", tx.SendTx())
+	})
+	app.Command("payload", "Create raw transaction", func(k *cli.Cmd) {
```
b00f, line 29: AS I checked your code, I am convinced that maybe we don't need the old style.

www/grpc/proto/zarb.proto
```diff
@@ -75,6 +75,7 @@ message BlockchainInfoRequest {
 
 message BlockchainInfoResponse {
   int64 height=1;
+  string stamp=2;
```
b00f, line 78: Use `Last_block_hash` instead. 
Stamp is not a blockchain thing. Stamp is a Tx related thing.

b00f (changes requested):
@d4rk5h0t 
Thanks for this PR. Sounds good. I think we might don't need `zarb tx payload` subcommand anymore. Let me check it. 

Another idea, maybe we can relocate `grpc.go` somewhere else like `util` folder. Because we can use this inside MainTest or other purposes later. Keep it in mind. Later we can discuss about it.

cmd/zarb/main.go
```diff
@@ -21,9 +22,13 @@ func zarb() *cli.Cli {
 		k.Command("verify", "Verify a signature", key.Verify())
 		k.Command("change-auth", "Change the passphrase of a keyfile", key.ChangeAuth())
 	})
-	app.Command("tx", "Create raw transaction", func(k *cli.Cmd) {
-		k.Command("bond", "Generate raw bond transaction", tx.BondTx())
-		k.Command("send", "Generate raw send transaction", tx.SendTx())
+	app.Command("tx", "Create, Sing, Publish transactions", func(k *cli.Cmd) {
```
b00f, line 25: `Create, sign and publish a transaction` is better. 
sign or sing ;)

cmd/zarb/main.go
```diff
@@ -21,9 +22,13 @@ func zarb() *cli.Cli {
...
+		k.Command("bond", "Send signed Bond transaction to network", tx.BondTx())
```
b00f, line 26: `Create, sign and publish a bond transaction`, may be better

cmd/zarb/main.go
```diff
@@ -21,9 +22,13 @@ func zarb() *cli.Cli {
...
+		k.Command("send", "Send signed Send transaction to network", tx.SendTx())
```
b00f, line 27: `Create, sign and publish a send transaction`, may be better

cmd/zarb/payload/send_tx.go
```diff
@@ -0,0 +1,121 @@
...
+			// ---
+			if *seqOpt == 0 {
+				cmd.PrintWarnMsg("Sequence number is not defined.")
+				c.PrintHelp()
+				return
+			}
+			seq = *seqOpt
+
+			if *amountOpt == 0 {
+				cmd.PrintWarnMsg("Stake is not defined.")
```
b00f, line 69: "Amount is not defined."

b00f:
I think we can remove `fee` option. Only show the last amount+fee that user is going to pay.

cmd/zarb/tx/common.go
```diff
@@ -0,0 +1,36 @@
...
+	k, err := key.DecryptKeyFile(keyfile, auth)
+	if err != nil {
+		cmd.PrintErrorMsg("Couldn't retrieve Key: %v", err)
+		return
+	}
+	k.ToSigner().SignMsg(trx)
+
+	cmd.PrintWarnMsg("you are about to publish:")
+	cmd.PrintJSONObject(trx)
+	confirm := cmd.PromptInput("are you sure [y/n]:")
```
b00f, line 23:
```
confirm := cmd.PromptInput("The operation is not reversible. Are you sure? [yes/no]")
	if !strings.HasPrefix(strings.ToLower(confirm), "yes") {
```

b00f: This is really important. Let user write `yes`. Make sure not accidentally press a `y` key.

cmd/zarb/tx/common.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/zarbchain/zarb-go/keystore/key"
+	"github.com/zarbchain/zarb-go/tx"
+	zarb "github.com/zarbchain/zarb-go/www/grpc/proto"
+)
+
+func signAndPublish(trx *tx.Tx, keyfile, auth string, grpcClient zarb.ZarbClient) {
+	//sign transaction
+	k, err := key.DecryptKeyFile(keyfile, auth)
+	if err != nil {
+		cmd.PrintErrorMsg("Couldn't retrieve Key: %v", err)
```
b00f, line 16: `Couldn't retrieve the key`

cmd/zarb/tx/common.go
```diff
@@ -0,0 +1,36 @@
...
+		return
+	}
+	k.ToSigner().SignMsg(trx)
+
+	cmd.PrintWarnMsg("you are about to publish:")
+	cmd.PrintJSONObject(trx)
```
b00f, line 22: better print the raw bytes for user that want to send the raw transaction.

use `cmd.PrintLine()` for line separator.

cmd/zarb/tx/bond_tx.go
```diff
@@ -112,10 +110,52 @@ func BondTx() func(c *cli.Cmd) {
...
+
+			grpcClient, err := cmd.GetRPCClient(rpc)
+			if err != nil {
+				cmd.PrintErrorMsg("Couldn't connect to RPC Server: %v", err)
+				return
+			}
+
+			seq = *seqOpt
+			if *seqOpt == 0 {
+				seq, err = cmd.GetSequence(grpcClient, bonder)
```
b00f, line 140: I have a suggestion here. 
If you make `func grpcClient(grpcOpt *string) {... }`
Then you only ask user anytime you want to query or broadcast the network. 
`cmd.GetSequence(grpcClient(grpcOpt), bonder)`

So if user wants to set all the options manually (offline), no need to ask gRPC endpoint.

cmd/zarb/tx/common.go
```diff
@@ -0,0 +1,36 @@
...
+		cmd.PrintWarnMsg("Opration aborted!")
+		return
+	}
+
+	// publish
+	signedTrx, _ := trx.Encode()
+	if id, err := cmd.SendTx(grpcClient, signedTrx); err != nil {
+		cmd.PrintErrorMsg("Couldn't publish transaction: %v", err)
+	} else {
+		cmd.PrintSuccessMsg("transaction sent with Id: %v", id)
```
b00f, line 34: "Transaction sent with ID"

b00f:
I wish we could write test for this part. because it is really important. Later we can discuss about how to write test for CLI.

d4rk5h0t:
> 
> 
> @d4rk5h0t
> Thanks for this PR. Sounds good. I think we might don't need `zarb tx payload` subcommand anymore. Let me check it.
> 
> Another idea, maybe we can relocate `grpc.go` somewhere else like `util` folder. Because we can use this inside MainTest or other purposes later. Keep it in mind. Later we can discuss about it.

sure but it wasn't me 😂

cmd/gRPC.go
Ali-Meh: sure moved it to util/grpc.go

cmd/zarb/main.go
Ali-Meh: we might need them for smart-contracts implementations so i suggest keeping them

www/grpc/proto/zarb.proto
Ali-Meh: sure applyed

cmd/zarb/main.go
Ali-Meh: you right my bad

cmd/zarb/main.go
Ali-Meh: done

Ali-Meh:
> I wish we could write test for this part. because it is really important. Later we can discuss about how to write test for CLI.

@b00f 
I agree with you, Actually I tried to implement some tests for CLI but I realized it's heavily relies on tested modules like gRPC or transaction so may be pushing it much.
besides it needs some mocks to implement if we could use some libraries to generate mocks, it would be much easier but it may have inconsistency with current code base

cmd/zarb/main.go
b00f: This is redundant to me now. With the new commands, we can still make transactions and sign them off line. 
Please consider that users might raise issues about them. Let archive them now.

cmd/zarb/tx/bond_tx.go
d4rk5h0t: @b00f
I tried to implement it this idea.
let me know if you got better way

Ali-Meh: i tried to implement this idea let me know if you got better way

cmd/zarb/tx/bond_tx.go
```diff
@@ -112,10 +110,40 @@ func BondTx() func(c *cli.Cmd) {
...
+			}
+
+			stamp, err = crypto.HashFromString(*stampOpt)
+			if err != nil {
+				stamp, err = util.GetStamp(promptRPCEndpoint(*grpcOpt))
+				if err != nil {
+					cmd.PrintErrorMsg("Couldn't retrieve stamp from RPC Server: %v", err)
+					return
+				}
+			}
```
b00f, line 142: This part has a potential problem. User may enter invalid hash. So we ignore it and query from the node. 
Maybe we can rewrite it like this:
```
			
			if stampOpt == nil {
                stamp, err = crypto.HashFromString(*stampOpt)
                if err != nil {
					cmd.PrintErrorMsg("Stamp is not valid: %v", err)
					return
				}
           } else { 
				stamp, err = util.GetStamp(promptRPCEndpoint(*grpcOpt))
				if err != nil {
					cmd.PrintErrorMsg("Couldn't retrieve stamp from RPC Server: %v", err)
					return
				}
			}
```

cmd/zarb/tx/bond_tx.go
```diff
@@ -112,10 +110,40 @@ func BondTx() func(c *cli.Cmd) {
...
+
+			//RPC
+			seq = *seqOpt
+			if *seqOpt == 0 {
+				seq, err = util.GetSequence(promptRPCEndpoint(*grpcOpt), bonder)
+				if err != nil {
+					cmd.PrintErrorMsg("Couldn't retrieve Sequence number from RPC Server: %v", err)
+					return
+				}
+			}
```
b00f, line 133: How about this:
```
        if seqOpt != nil {
                seq = *seqOpt
           } else { 
				seq, err = util.GetSequence(promptRPCEndpoint(*grpcOpt), bonder)
				if err != nil {
					cmd.PrintErrorMsg("Couldn't retrieve Sequence number from RPC Server: %v", err)
					return
				}
			}
```

account/account.go
```diff
@@ -72,8 +72,9 @@ func (acc *Account) UnmarshalJSON(bs []byte) error {
 }
 
 func (acc Account) Fingerprint() string {
-	return fmt.Sprintf("{ %s %v}",
+	return fmt.Sprintf("{ %s %d %v}",
```
b00f, line 75: Could you please update Fingerprint for validator as well?

Ali-Meh: sure why not

b00f (approved):

Issue #209, opened 2021-05-14
WWW: Refactor gRPC API Messages


d4rk5h0t:
### Describe the problem you'd like to have solved

current implementation of gRPC server consist of 2 API message types which makes interaction with the API a bit harder

### Describe the ideal solution

make all the gRPC APIs return and receive same structures 
It may be better to define distinct object structures to be return from APIs but It might make querying server harder but since RPC server isn't query service I think its better to take care of queries with explorer services or wrap it on GraphQL service and have uniform APIs on RPC side
## Alternatives and current work-arounds

refactor APIs to return same structures types as example of this issue in [PR 204](https://github.com/zarbchain/zarb-go/pull/204/files?file-filters%5B%5D=.go&file-filters%5B%5D=.json&file-filters%5B%5D=.proto#diff-3be930d36fe301f565030d3d1faf0ce8d50dc958a4d83cd5ed45a8e4bfa4b406R17-R37) there are 2 different implementations of client which are fairly similar

Ali-Meh:
I agree And I can get it done but which way we gonna take
we need all APIs feel same either way I can refactor all to verbose thing or defined structs

b00f:
Yes. You are right. Right now we have two different approaches for gRPC which is a bit confusing. 
For sake of simplicity we can think about different output level. 
For example, 'getBlock' method: 
- with verbosity 0 just returns block data. 
- with verbosity 1 returns block information 
- with verbosity 2 returns block and transaction information 

This is almost same as getBlock RPC method in [Bitcoin](https://developer.bitcoin.org/reference/rpc/getblock.html)

Ali-Meh:
> For example, 'getBlock' method:
> 
>     * with verbosity 0 just returns block data.
> 
>     * with verbosity 1 returns block information
> 
>     * with verbosity 2 returns block and transaction information

@b00f 
there is no doubt we might need this kind of verbosity levels in gRPC we can achieve this by not filling sub structs or messages but right now there is no struct to begin with 

we first need to decide if we want to have solid predefined object messages or dynamic base64 encoded json messages

b00f:
@Ali-Meh @d4rk5h0t  I think we can close this issue now.

PR #223, opened 2021-05-23
www: gRPC refactor

Ali-Meh:
## Description
it will refactor gRPC service apis and messages 

## Related issue(s)
#209 

## Checklist

- [x] I have added tests to cover my changes.
- [x] CHANGELOG is updated.


www/grpc/proto/zarb.proto
```diff
@@ -126,4 +120,51 @@ message Peer{
...
+
+message BlockInfo {
+  int32 version = 1;
+  string last_block_hash = 2;
+  string state_hash=3;          
+  string tx_ids_hash=4;
+  string last_certificate_hash=5;
+  bytes sortition_seed=6;
+  string proposer_address=7;   
+  int64 round=8;
```
b00f, line 140: int32 is better for round.

www/grpc/proto/zarb.proto
```diff
@@ -126,4 +120,51 @@ message Peer{
...
+enum PayloadType {
+  SEND_PAYLOAD=0;
+  BOND_PAYLOAD =1;
+  SORTITION_PAYLOAD=2;
+  UNBOND_PAYLOAD =3;
+}
+
+enum BlockVerbosity {
+  BLOCK_HASH = 0;
+  BLOCK_DATA = 1;
```
b00f, line 168: Suggestion: renaming "BLOCK_DATA" to "BLOCK_INFO", because block data sounds like raw block data bytes.

www/grpc/proto/zarb.proto
```diff
@@ -126,4 +120,51 @@ message Peer{
...
+  repeated int32 committers=9;
+  repeated int32 absentees=10;
+  string signature=11;
+}
+
+message Transaction{
+  string id=1;    
+  int32 Version =2;  
+  string stamp=3;    
+  int64 sequence=4;
```
b00f, line 150: Since sequence in account & validator are almost same as sequence in transaction, type also should be the same. 
here type in `int64` but in account & validator is `int32`.

www/grpc/proto/zarb.proto
```diff
@@ -126,4 +120,51 @@ message Peer{
   int32 received_messages = 7;
   int32 invalid_messages = 8;
   int32 received_bytes = 9;
+}
+
+message Account{
```
b00f, line 125: Let's have harmony here. All structure names ends with `Info` 

Like: `AccountInfo`, `BlockInfo`, `ValidatorInfo`, `TransactionInfo`

www/grpc/transaction.go
```diff
@@ -22,25 +22,20 @@ func (zs *zarbServer) GetTransaction(ctx context.Context, request *zarb.Transact
...
-
-	return res, nil
+	return &zarb.TransactionResponse{
+		Tranaction: &zarb.Transaction{
+			Id:        trx.ID().String(),
+			Version:   int32(trx.Version()),
+			Stamp:     trx.Stamp().String(),
+			Sequence:  int64(trx.Sequence()),
+			Fee:       trx.Fee(),
+			Type:      zarb.PayloadType(trx.PayloadType() - 1), //enums starting from 0
```
b00f, line 32: This logic is not helpful. We have two different numbers for the same type.
Do really need to report type to user? If yes, let report the actual amount.

www/grpc/transaction.go
```diff
@@ -22,25 +22,20 @@ func (zs *zarbServer) GetTransaction(ctx context.Context, request *zarb.Transact
...
+			Payload:   trx.Payload().Signer().RawBytes(),
```
b00f, line 33: Payload can be a [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) field. Therefore we can decode the payload.

b00f (changes requested):

www/grpc/transaction.go
d4rk5h0t: @b00f 
`oneof` field as your mentioned  clarifies payload content but then we face indexing fields of Protobuf then we might need to reserve some fields or  start using high indexes like 30 - 40 for these payloads

Ali-Meh: @b00f 
yeah sure `oneof` field will clarify the content of payload but then we will face the indexing issue we will need to reserve bunch of fields or start indexing payload high like from 30 or so

www/grpc/transaction.go
Ali-Meh: @b00f 
if we take `oneof` approach we definitely don't need it for dynamic type languages but not sure about static type languages though

b00f (approved):

PR #226, opened 2021-06-08
Replace sync.RWMutux with deadlock.RWMutux Only for test

d4rk5h0t:
## Description

replace Replace sync.RWMutux with deadlock.RWMutux only in test env
> Warning: its not the most scalable solution but it does for now we might better wrap `RWMutux` in a package controlled by environment variable to use ether `sync` or `deadlock` pakages

imports of sync package as described below will not get replaced
```go
sync "sync"
```
* as you can see in file [network/network.go](https://github.com/zarbchain/zarb-go/pull/226/files#diff-96a7f5c352cd3d2d1ee3c7c55ed95e668c6edb28a4fa337e6c60270471bdab61) there is 2 sync packages imported `sync` and `syncer` with latter one won't get replaced in test env 

## Related issue(s)
Some Performance Upgrade

## Checklist

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,9 +12,6 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v2
-      
-    - name: Replace sync.RWMutux with deadlock.RWMutux
-      run: make replace_mutux
 
     - name: Test with coverage
       run: go test -gcflags=-l -v  -coverprofile=coverage.txt -covermode=atomic ./...
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,6 +12,9 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v2
+      
+    - name: Replace sync.RWMutux with deadlock.RWMutux
+      run: make replace_mutux
 
     - name: Test with coverage
       run: go test -gcflags=-l -v  -coverprofile=coverage.txt -covermode=atomic ./...
--- a/.github/workflows/race.yml
+++ b/.github/workflows/race.yml
@@ -13,5 +13,8 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    - name: Replace sync.RWMutux with deadlock.RWMutux
+      run: make replace_mutux
+
     - name: Test with data race detection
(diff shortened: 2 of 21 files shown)


Makefile
```diff
@@ -78,6 +78,8 @@ fmt: herumi
 		--enable=misspell \
 		--enable=gosec
 
+replace_mutux:
```
b00f, line 81: I think we don't need a new target here. Let's don't change the makefile. Better add a new workflow for only checking data race when user send a pull request. The report is enough for finding the data race condition.

.github/workflows/race.yml
```diff
@@ -13,5 +13,8 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    - name: Replace sync.RWMutux with deadlock.RWMutux
```
b00f, line 16: Since the tests are the same for all workflows, I suggest to change this workflow (race.yml) to find the deadlocks situations. 
Let's remove the `replace_mutux` from the Makefile and add it here.

.github/workflows/coverage.yml
```diff
@@ -12,6 +12,9 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v2
+      
+    - name: Replace sync.RWMutux with deadlock.RWMutux
+      run: make replace_mutux
```
b00f, line 17: Revert this.

.github/workflows/testing.yml
```diff
@@ -16,6 +16,9 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v2
+    
+    - name: Replace sync.RWMutux with deadlock.RWMutux
```
b00f, line 20: Revert this also

.github/workflows/race.yml
b00f: line # 17: `find . -type f -name "*.go" -not -path '*/\.*' -exec sed -i -- 's/\t"sync"/\tsync "github.com\/sasha-s\/go-deadlock"/g' {} +`

Line #2: `Deadlock and data race detection`

Rename the file to `deadlock.yml`

Makefile
b00f, line 81: Revert this also

b00f (approved):

PR #229, opened 2021-11-17
Define interface for crypto

b00f:
## Description

Defining a new trait for crypto stuff

## Checklist

- [ ] CHANGELOG is updated.


node/node_test.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	"github.com/zarbchain/zarb-go/crypto/hash"
 	"github.com/zarbchain/zarb-go/genesis"
 	"github.com/zarbchain/zarb-go/param"
 	"github.com/zarbchain/zarb-go/util"
 	"github.com/zarbchain/zarb-go/validator"
 )
 
 func TestRunningNode(t *testing.T) {
-	_, pb, pv := crypto.RandomKeyPair()
+	pub, pv := bls.RandomKeyPair()
```
Ali-Meh, line 20: pv => prv

state/state.go
```diff
@@ -272,7 +273,8 @@ func (st *state) ProposeBlock(round int) (*block.Block, error) {
 
 	stateHash := st.stateHash()
 	timestamp := st.proposeNextBlockTime()
-	newSortitionSeed := st.lastInfo.SortitionSeed().Generate(st.signer)
+	seed := st.lastInfo.SortitionSeed()
+	newSortitionSeed := seed.Generate(st.signer)
```
Ali-Meh, line 278: I don't see `seed` being used after here And I don't think other than consuming memory of seed variable it has other impact
here as well
https://github.com/zarbchain/zarb-go/pull/229/files#diff-ad95c17c31b539a3e385a23d86bbecf4d8a925ba6679aec2e66b5202ba4c4f81R350-R351

state/validation_test.go
```diff
@@ -6,12 +6,22 @@ import (
...
 )
 
+func aggregate(sigs []crypto.Signature) *bls.Signature {
+	blsSigs := make([]*bls.Signature, len(sigs))
+	for i, s := range sigs {
+		blsSigs[i] = s.(*bls.Signature)
+	}
+	return bls.Aggregate(blsSigs)
+}
+
```
Ali-Meh, line 24: nice But I think `crypto.Singature` it self be able to handle this rather than consumer to be aggregating

sync/message/payload/aleyk.go
```diff
@@ -5,19 +5,20 @@ import (
...
-	Moniker         string           `cbor:"5,keyasint"`
-	PublicKey       crypto.PublicKey `cbor:"6,keyasint"`
-	Height          int              `cbor:"7,keyasint"`
-	Flags           int              `cbor:"8,keyasint"`
+	ResponseTarget  peer.ID        `cbor:"1,keyasint"`
+	ResponseCode    ResponseCode   `cbor:"2,keyasint"`
+	ResponseMessage string         `cbor:"3,keyasint"`
+	NodeVersion     string         `cbor:"4,keyasint"`
+	Moniker         string         `cbor:"5,keyasint"`
+	PublicKey       *bls.PublicKey `cbor:"6,keyasint"`
```
Ali-Meh, line 19: this way we are tightly coupled to bls implementation of public key while I think you are trying to decouple it

Ali-Meh (commented):
I don't get the reason behind separating `bls` from `crypto` module but if you are trying to decouple it I think using bls implementation directly in the code base is wrong pass while we have `crypto` interfaces for that matter.
in test files it might be right to use bls concrete implementation while I think even there we better off with dummy, default or mock implementation or even use `crypto` to proxy to `bls` implementation rather than directly using it and barely introducing crypto interfaces.

I found the links below helpful about interfaces
https://www.integralist.co.uk/posts/go-interfaces/#interfaces-in-go
https://www.efekarakus.com/golang/2019/12/29/working-with-interfaces-in-go.html

crypto/bls/private_key_test.go
```diff
@@ -57,7 +57,7 @@ func TestPrivateKeyFromString(t *testing.T) {
 }
 
 func TestMarshalingEmptyPrivateKey(t *testing.T) {
-	pv1 := PrivateKey{}
+	pv1 := &PrivateKey{}
```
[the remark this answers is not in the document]
b00f: Don with thanks
,

crypto/bls/private_key_test.go
```diff
@@ -47,7 +47,7 @@ func TestPrivateKeyFromBytes(t *testing.T) {
 }
 
 func TestPrivateKeyFromString(t *testing.T) {
-	_, _, priv1 := GenerateTestKeyPair()
+	_, priv1 := GenerateTestKeyPair()
```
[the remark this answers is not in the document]
b00f: Done

block/header.go
```diff
@@ -17,26 +18,26 @@ type Header struct {
 type headerData struct {
 	Version             int            `cbor:"1,keyasint"`
 	UnixTime            int64          `cbor:"2,keyasint"`
-	LastBlockHash       crypto.Hash    `cbor:"3,keyasint"`
-	StateHash           crypto.Hash    `cbor:"4,keyasint"`
-	TxIDsHash           crypto.Hash    `cbor:"5,keyasint"`
-	LastCertificateHash crypto.Hash    `cbor:"6,keyasint"`
+	LastBlockHash       hash.Hash      `cbor:"3,keyasint"`
```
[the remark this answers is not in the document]
b00f: Done

sync/message/payload/aleyk.go
b00f: The problem here is we can't deserialize public Key interface. To deserialized public key we have two choices:
- Don't use interface (current approach)
- Adding type to the the struct and based of type we deserialize the public key.

state/validation_test.go
b00f: Unfortunately, Only BLS has this functionality. we can't use schnorr since it has nonce.

state/state.go
b00f: It should be, we passed the linting pipeline

b00f:
> I don't get the reason behind separating `bls` from `crypto` module but if you are trying to decouple it I think using bls implementation directly in the code base is wrong pass while we have `crypto` interfaces for that matter. in test files it might be right to use bls concrete implementation while I think even there we better off with dummy, default or mock implementation or even use `crypto` to proxy to `bls` implementation rather than directly using it and barely introducing crypto interfaces.
> 
> I found the links below helpful about interfaces https://www.integralist.co.uk/posts/go-interfaces/#interfaces-in-go https://www.efekarakus.com/golang/2019/12/29/working-with-interfaces-in-go.html https://blog.chewxy.com/2018/03/18/golang-interfaces/

Yes, the test looks to be not beautiful after this PR. The main reason behind this PR, is to give type to address. Now address in 20 bytes, we can add one more bytes (type bytes) to assign a type to an address. In future we can support another more algorithm  other than BLS.

crypto/crypto.go
```diff
@@ -1,40 +1 @@
 package crypto
```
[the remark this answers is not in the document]
b00f: I removed this file.

state/state.go
Ali-Meh: linter checks for unused variable while this `seed` variable is used in line 277 so the linter won't mind

sync/message/payload/aleyk.go
Ali-Meh: if we can't de/serialize the public key what are marshal functionalities for?
I think we should have and identifier to indicate the type of public key rather than directly using implementer(bls)
```
type PublicKey interface {
	RawBytes() []byte
	String() string
	MarshalJSON() ([]byte, error)
	MarshalCBOR() ([]byte, error)
	SanityCheck() error
	Verify(msg []byte, sig Signature) bool
	Address() Address
	EqualsTo(right PublicKey) bool
}
```

Ali-Meh (approved):

state/state.go
b00f: The `Generate` method use pointer to seed. This is why we define `seed` here

Ali-Meh (approved):

d4rk5h0t (approved):

PR #235, opened 2021-12-15
Adding type to address

b00f:
## Description

Adding type to the address

## Checklist

- [x] CHANGELOG is updated.


crypto/address.go
```diff
@@ -11,14 +11,18 @@ import (
...
-// `zc` + type + data
-// type is 1 for schnorr signatures
+// Address format:
+// `zc` + type + data + checksum
+// type is 1 for BLS signatures
 
 const (
-	addressSize           = 20
+	AddressTypeBLS byte = 1
+)
```
d4rk5h0t, line 20: maybe use `iota` instead of 1?!

d4rk5h0t (approved):
LGTM

crypto/address.go
b00f: Good idea. Right now we have only one type of address.

PR #240, opened 2022-01-05
chore: upgrade libp2p package

mhatvan:
## Description

- upgrade `go-libp2p` and its related libs
- update method calls due to `go-libp2p` version upgrade
- remove duplicate `sync` import in `network.go`
- run `go mod tidy`
- update `go` version in `testing.yml` to 1.17
- update `docker` container to use `alpine` 3.15

Please note that `lp2pdiscovery` is updated to use the `mdns_legacy` module which keeps compatibility with the current code.
There is also `mdns` which should probably be used instead when refactoring the code. 

## Related issue(s)

Closes https://github.com/zarbchain/zarb-go/issues/238

## Checklist

- [ ] CHANGELOG is updated.


mhatvan:
`make unit_test` is passing for me locally, but fails in the CI, any idea why?

`make docker` is probably failing due to https://github.com/lucas-clemente/quic-go only supporting go 1.16 and higher I assume, while the docker image is using go 1.15

b00f:
@mhatvan 

You can update [alpine](https://hub.docker.com/_/alpine) version inside [dockerfile](https://github.com/zarbchain/zarb-go/blob/main/containers/Dockerfile) to support go version 17:
```
FROM alpine:3.15 as builder

RUN apk add --no-cache git go gmp-dev build-base g++ openssl-dev
RUN go version
...
```

About the MDNS legacy version, they have plan to remove it. Look at here: https://github.com/libp2p/go-libp2p/pull/1192/files
It's better to not use it anymore.

> I probably caused these incompatibilities by running go mod tidy?

Shouldn't be. It's really weird, probably related to previous issue.

mhatvan:
> About the mDNS legacy version, they have plan to remove it. Look at here: libp2p/go-libp2p#1192 (files)
It's better to not use it anymore.

I tried to update the respective code but don't know how exactly. I would hand this over to you.

b00f (approved):

PR #250, opened 2022-02-20
Update the block header and previous certificate in grpc blockinfo

nagarajmanjunath:
## Description

* Fix gRPC proto.
* Updated the block info response with block header and previous hash certificate.
* Fix the test cases.

## Checklist

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] CHANGELOG is updated.


www/grpc/proto/zarb.pb.go
```diff
@@ -1,12 +1,13 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.27.1
-// 	protoc        v3.17.3
+// 	protoc-gen-go v1.25.0
```
b00f, line 3: I think your protoc version is old.

b00f (changes requested):

www/grpc/proto/zarb.pb.go
nagarajmanjunath: I ll update it.

b00f (approved):

PR #427, opened 2022-11-16
Convert HTTP module to use GRPC module instead of CAPNP module

### LINKED ISSUE #373: Adding Withdraw payload for gRPC
Withdraw payload is missed in gRPC proto file.

### LINKED ISSUE #374: Adding Unbod payload for gRPC
Unbond payload is missed in proto file

### LINKED ISSUE #401: Peer ID is not in String format
## Description

image

In HTTP service peer ID is not in string format

### How To Reproduce

Just run node locally or check this url: http://172.104.46.145:8080/network

### What Happened

Peer ID should be in string format. like "12D3KooWNYD4bB82YZRXv6oNyYPwc5ozabx2epv75ATV3D8VD3Mq"

### LINKED ISSUE #418: Replacing CapnP with gRPC for http package
CapnP package is deprecated and should be removed. We need to replace CapnP with gRPC in http package:
https://github.com/pactus-project/pactus/blob/58c9045d1b0bc195ab91fc96e5aeb5fa9bb84e6e/www/http/server.go#L24

braewoods:
## Description

This converts the HTTP module to use the GRPC module instead of the CAPNP module. This also required some more extensions to the GRPC module as there was some missing information. Since this was the last user of the CAPNP module, it is also removed.

## Related issue(s)

This should close #418. It should also close #401 since the output was fixed during the conversion. And it should also close #373 and close #374 since I added the missing payload types to GRPC module to finish this conversion.

## Checklist

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] CHANGELOG is updated.


go.mod
```diff
@@ -27,10 +27,8 @@ require (
 	github.com/tyler-smith/go-bip39 v1.1.0
 	go.nanomsg.org/mangos/v3 v3.4.2
 	golang.org/x/crypto v0.0.0-20220824171710-5757bc0c5503
-	golang.org/x/net v0.0.0-20220812174116-3211cb980234
 	google.golang.org/grpc v1.49.0
 	google.golang.org/protobuf v1.28.1
-	zombiezen.com/go/capnproto2 v2.18.2+incompatible
```
b00f, line 33: Two less dependency! Awesome!

tests/main_test.go
```diff
@@ -1,165 +0,0 @@
-package tests
```
b00f, line 1: This test is kind of integration or system test. I see you removed it. We can keep these test and just use gRPC for querying the node.

tests/main_test.go
```diff
@@ -1,165 +0,0 @@
...
-	"github.com/pactus-project/pactus/www/capnp"
-	"zombiezen.com/go/capnproto2/rpc"
-)
-
-var tSigners []crypto.Signer
-var tConfigs []*config.Config
-var tNodes []*node.Node
-var tCapnpAddress = "0.0.0.0:1337"
-var tGenDoc *genesis.Genesis
-var tCapnpServer capnp.PactusServer
```
b00f, line 31: Replace this with a gRPC server, then we can have this test

www/grpc/proto/blockchain.proto
```diff
@@ -82,13 +82,15 @@ message ValidatorInfo {
   uint32 last_joined_height = 6;
   uint32 unbonding_height = 7;
   string address = 8;
+  bytes hash = 9;
```
b00f, line 85: Please let hash be the first field in the account. Also include data field.  
```proto
message ValidatorInfo {
  string hash = 1;
  int32 data = 2;
  <other fields come here>
}
```

www/grpc/proto/blockchain.proto
```diff
@@ -104,6 +106,7 @@ message CertificateInfo {
   repeated int32 committers = 2;
   repeated int32 absentees = 3;
   bytes signature = 4;
+  bytes hash = 5;
```
b00f, line 109: Same here. Certificate data should be also reported to the user.

www/grpc/proto/blockchain.proto
```diff
@@ -82,13 +82,15 @@ message ValidatorInfo {
...
   string address = 8;
+  bytes hash = 9;
 }
 
 message AccountInfo {
   string address = 1;
   int32 number = 2;
   int32 sequence = 3;
   int64 Balance = 4;
+  bytes hash = 5;
```
b00f, line 93: same here

www/http/block.go
```diff
@@ -21,62 +21,55 @@ func (s *Server) GetBlockByHeightHandler(w http.ResponseWriter, r *http.Request)
...
 		return
 	}
 
-	seed := b.Header().SortitionSeed()
+	s.blockByHeight(w, res.Height)
+}
+
+func (s *Server) blockByHeight(w http.ResponseWriter, blockHeight uint32) {
+	res, err := s.blockchain.GetBlock(s.ctx,
+		&pactus.GetBlockRequest{Height: blockHeight, Verbosity: 0})
```
b00f, line 43:
```suggestion
		&pactus.GetBlockRequest{Height: blockHeight, Verbosity: pactus.BlockVerbosity_BLOCK_TRANSACTIONS)})
```

www/http/server.go
```diff
@@ -34,18 +38,26 @@ func NewServer(conf *Config) *Server {
...
 }
 
-func (s *Server) StartServer(capnpServer string) error {
+func (s *Server) StartServer(grpcServer string) error {
 	if !s.config.Enable {
 		return nil
 	}
 
-	c, err := net.Dial("tcp", capnpServer)
+	c, err := grpc.DialContext(
```
b00f, line 46:
```suggestion
	conn, err := grpc.DialContext(
```

I suggest give a better name to variable `c`

b00f (changes requested):

b00f:
@braewoods Thanks. All looks good. JUst the `tests` folder is removed. That test is really important to make sure the system works fine

tests/main_test.go
braewoods: > @braewoods Thanks. All looks good. JUst the `tests` folder is removed. That test is really important to make sure the system works fine

I thought they were exclusive to CAPNP due to how tightly integrated to it they seemed to be which is why they were removed.

braewoods:
> @braewoods Thanks. All looks good. JUst the `tests` folder is removed. That test is really important to make sure the system works fine

I thought they were exclusive to CAPNP so that's why it was removed.

www/http/server.go
braewoods: Integrated into next force push.

braewoods:
It should now fix all the discussed issues.

b00f (approved):

PR #437, opened 2023-02-28
feat!: indexing public key

b00f:
## Description

This PR indexes the public key by addresses and removes the public key from the transactions if it is already known. 
This could potentially reduce the size of the blocks, since some transactions may already have known public keys.

BREAKING CHANGE

### DIFF
--- a/store/interface.go
+++ b/store/interface.go
@@ -1,6 +1,8 @@
 package store
 
 import (
+	"bytes"
+
 	"github.com/pactus-project/pactus/crypto"
 	"github.com/pactus-project/pactus/crypto/hash"
 	"github.com/pactus-project/pactus/types/account"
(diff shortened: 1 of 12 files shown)


store/interface.go
```diff
@@ -33,37 +35,44 @@ func (s *StoredBlock) ToBlock() *block.Block {
...
+	trx := new(tx.Tx)
+	r := bytes.NewReader(s.Data)
+	if err := trx.Decode(r); err != nil {
 		panic(err)
 	}
-	if !trx.ID().EqualsTo(s.TxID) {
-		panic("invalid data. transaction id does not match")
-	}
+
+	// TODO: we can set public key, if it doesn't set
```
b00f, line 52: 1- move this to sync
2- no panic

Issue #472, opened 2023-05-18
password option for start command


b00f:
## Describe the problem you'd like to have solved

User in Discord requested this feature:

> How can I ran the node with system service? The node has no parameters to pass the wallet password or config file.

### Describe the ideal solution

We need to let user to set `password` as option for the start command in `daemon` and `gui`.

### Alternatives and current work-around

There is no workaround.

PR #473, opened 2023-05-18
fix: adding password option to the start commands

b00f:
## Description

The password option has been added to the `start` command for both `daemon` and `gui`. 
To prevent any possible mistakes during the Mainnet launch, the working directory for the Testnet has been changed to `~/pactus/testnet`.

Fixing #472

### DIFF
--- a/consensus/log/messages.go
+++ b/consensus/log/messages.go
@@ -49,6 +49,6 @@ func (m *Messages) voteSet(voteType vote.Type) *voteset.VoteSet {
 		return m.changeProposerVotes
 	}
 
-	logger.Panic("unexpected vote type %d", voteType)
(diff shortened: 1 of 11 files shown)


consensus/log/messages.go
```diff
@@ -49,6 +49,6 @@ func (m *Messages) voteSet(voteType vote.Type) *voteset.VoteSet {
 		return m.changeProposerVotes
 	}
 
-	logger.Panic("unexpected vote type %d", voteType)
```
themantre, line 52: Is it safe panic here?

b00f: We checked the vote_type parameter in the `SanityCheck()` function. 
Returning an error at this point would require multiple error checks in various other locations.

themantre (approved):

Issue #478, opened 2023-05-20
Reset config file to default when user deletes it


b00f:
## Description

When the configuration file is deleted, it prevents the node from running.

### How To Reproduce

1- Delete the configuration file.
2- Run the node again.

### What Happened

The node fails to start and returns the following error:
`open config.toml: no such file or directory`

### Expected Behavior

The system should automatically restore the configuration file to its default state when it's detected or missing.

b00f:
Restoring the configuration file when it is deleted is challenging. 
One such challenge is the inability to determine the exact number of validators a user has created, which is necessary for restoring the default configuration.

However, we have encountered similar situations where the config file becomes obsolete or contains mismatched fields, especially when a new version of the software is released. In issue #864, we addressed this by creating a backup of the current config file and loading it in non-strict mode. By doing so, we were able to retrieve the number of validators and restore the config file to its default values.

Ja7ad:
@b00f I have an ideal solution:

1. When the service is started, if the config file is loaded, it creates a file next to it named `.config.toml.save` with the current settings that have been loaded.

2. If the `config.toml` file is not found, attempt to recover it from the `.config.toml.save` file.

3. If a change is made to the `config.toml` file, update the `.config.toml.save` file.

4. If neither file exists, throw an error.

b00f:
@Ja7ad 
This makes loading the configuration file more complex. 

I am interested in seeing the issues or problems that users might encounter in action. 
Once we have enough reports, we can make better decisions.

PR #484, opened 2023-05-24
fix: restoring config file to the default

b00f:
## Description

This pull request fixes the issue with restoring the configuration file to the default values when unable to load the config.

## Related issue(s)

Related to #478

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -333,18 +333,20 @@ func StartNode(workingDir string, passwordFetcher func(*wallet.Wallet) (string, 
 		PrintWarnMsg("Unable to load the config: %s", err)
 		PrintInfoMsg("Attempting to restore the config to the default values...")
 
-		// Let's backup the config first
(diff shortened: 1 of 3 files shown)


cmd/cmd.go
```diff
@@ -326,9 +327,36 @@ func StartNode(workingDir string, passwordFetcher func(*wallet.Wallet) (string,
...
+			return nil, nil, err
+		}
+
+		confBackupPath := fmt.Sprintf("%v_bak_%s", confPath, time.Now().Format("2006_01_02"))
+		err = os.Rename(confPath, confBackupPath)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		switch gen.ChainType() {
```
themantre, line 348: Add comment here, Now, we try to restore the config file with the nbumer of validators from old config

themantre (changes requested):

themantre (approved):

PR #486, opened 2023-05-26
fix: waiting for proposal in pre-commit phase

b00f:
## Description

This pull request addresses a situation observed in the Testnet where certain nodes with slow connections become absentees in the certificate. This occurs when they receive block announcements before entering the pre-commit phase, possibly due to late proposals or votes.

To address this issue, the following changes have been made:

- When there are sufficient pre-commit votes but no proposal, the system now waits to receive the proposal before proceeding.
- If a proposal is received after committing a block, it is now accepted.
- If timeout expires in pre-commit phase, we move to the change-proposer phase.

These scenarios are covered by tests, and  also been tested using [TLA+](https://lamport.azurewebsites.net/tla/tla.html) to ensure its correctness.


consensus/consensus.go
```diff
@@ -183,15 +178,26 @@ func (cs *consensus) SetProposal(p *proposal.Proposal) {
...
-	if err := cs.state.ValidateBlock(p.Block()); err != nil {
-		cs.logger.Warn("invalid block", "proposal", p, "err", err)
-		return
+	if p.Height() == cs.state.LastBlockHeight() {
+		// A slow node might receive a proposal after committing the proposed block.
+		// In this case, we accept the proposal and allow nodes to continue.
+		// By doing so, we enable the validator to broadcast its votes and
+		//prevent it from being marked as absent in the block certificate.
+		cs.logger.Trace("block is committed for this height", "proposal", p, "block hash", cs.state.LastBlockHash())
+		if p.Block().Hash() != cs.state.LastBlockHash() {
```
themantre, line 187: add warning log here

consensus/precommit.go
```diff
@@ -23,7 +23,9 @@ func (s *precommitState) decide() {
 	precommitQH := precommits.QuorumHash()
 	if precommitQH != nil {
 		s.logger.Debug("precommit has quorum", "precommitQH", precommitQH)
-		s.enterNewState(s.commitState)
+		if s.hasVoted {
```
themantre, line 26: query proposal here. (don't forget to add a log)

themantre (changes requested):

consensus/consensus.go
b00f: done

consensus/precommit.go
b00f: It already called `vote()` before and should be queried

consensus/consensus.go
```diff
@@ -183,15 +178,27 @@ func (cs *consensus) SetProposal(p *proposal.Proposal) {
...
-	}
-
-	if err := cs.state.ValidateBlock(p.Block()); err != nil {
-		cs.logger.Warn("invalid block", "proposal", p, "err", err)
-		return
+	if p.Height() == cs.state.LastBlockHeight() {
+		// A slow node might receive a proposal after committing the proposed block.
+		// In this case, we accept the proposal and allow nodes to continue.
+		// By doing so, we enable the validator to broadcast its votes and
+		//prevent it from being marked as absent in the block certificate.
```
themantre, line 185:
```suggestion
		// prevent it from being marked as absent in the block certificate.
```

themantre (approved):

PR #504, opened 2023-06-09
feat!: updating sortition executor

b00f:
## Description

This PR introduces changes to the sortition executor to address several issues:

- It fixes a bug that was causing duplicate sortition transactions.
- It corrects the calculation used to determine whether a proposer has proposed a block or not.

BREAKING CHANGE


sandbox/sandbox.go
```diff
@@ -18,23 +18,24 @@ import (
...
-	validators      map[crypto.Address]*ValidatorStatus
+	accounts        map[crypto.Address]*accountStatus
+	validators      map[crypto.Address]*validatorStatus
 	params          param.Params
 	totalAccounts   int32
 	totalValidators int32
 }
 
-type ValidatorStatus struct {
+type validatorStatus struct {
```
b00f, line 32: rename them to `sandboxAccount` and `sandoxValidator`

PR #507, opened 2023-06-13
fix: updating linkedmap to use generic

nagarajmanjunath:
## Description

* updating linked map to use generic


sandbox/sandbox.go
```diff
@@ -74,8 +74,8 @@ func (sb *sandbox) shouldPanicForUnknownAddress() {
 }
 
 func (sb *sandbox) Account(addr crypto.Address) *account.Account {
-	sb.lk.RLock()
-	defer sb.lk.RUnlock()
+	sb.lk.Lock()
```
b00f, line 77: This should be RLock

b00f (approved):

PR #509, opened 2023-06-14
fix: get recent blocks by stamp

nagarajmanjunath:
## Description

This PR makes it possible to fetch the latest block using the stamp, which removes the need to search the levelDB for the specific block. 
This improve the performance of executing sortition transactions that need to obtain the sortition seeds based on the block stamp.

### DIFF
--- a/store/mock.go
+++ b/store/mock.go
@@ -151,9 +151,6 @@ func (m *MockStore) LastCertificate() (uint32, *block.Certificate) {
 	return m.LastHeight, m.LastCert
 }
 func (m *MockStore) RecentBlockByStamp(stamp hash.Stamp) (uint32, *block.Block) {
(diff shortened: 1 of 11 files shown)


store/mock.go
```diff
@@ -150,29 +150,17 @@ func (m *MockStore) LastCertificate() (uint32, *block.Certificate) {
 	}
 	return m.LastHeight, m.LastCert
 }
-func (m *MockStore) FindBlockHashByStamp(stamp hash.Stamp) (hash.Hash, bool) {
+func (m *MockStore) RecentBlockByStamp(stamp hash.Stamp) (uint32, *block.Block) {
 	if stamp.EqualsTo(hash.UndefHash.Stamp()) {
```
b00f, line 154: This check is not necessary. Remove it

b00f (approved):

PR #512, opened 2023-06-15
refactor: GetValidators return all validators in state validators map

### LINKED ISSUE #502: gRPC: GetValidators should return the list of all validator addresses
## Description

The `GetValidators` API in gRPC is returning only the validators inside the committee. However, it should return the list of all validator addresses in the blockchain.

### How To Reproduce

Call the `GetValidators` API, for example at this node:  [/v1/blockchain/validators](http://172.104.46.145/#/Blockchain/Blockchain_GetValidators)

### What Happened

It returns the list of validators inside the committee. 
This information can already be obtained from the `GetBlockchainInfo` API.

### Expected Behavior

The expected behavior is for the `GetValidators` API to return the list of all validators in the blockchain. We have a [map of all validators](https://github.com/pactus-project/pactus/blob/7fd42b2767110fed6cab20c4696f36179782a51d/store/validator.go#L15) in memory within the store. To achieve this, we need to update the state facade to return all validator addresses (not objects). Then, in the gRPC server, we can retrieve the validator addresses from the state.

**UPDATE**

Instead of returning all validator objects, it would be better to only retrieve a list of validator addresses. This approach is more efficient, especially when dealing with a large number of validators, such as over 1000. Returning all validator objects could potentially cause delays in the response time.

Ja7ad:
## Description

gRPC: GetValidators should return the list of all validator addresses 

## Related issue(s)

Fixed #502

/claim #502


state/facade.go
```diff
@@ -45,6 +45,7 @@ type Facade interface {
 	AccountByNumber(number int32) *account.Account
 	ValidatorByAddress(addr crypto.Address) *validator.Validator
 	ValidatorByNumber(number int32) *validator.Validator
+	Validators() []*validator.Validator
```
b00f, line 48: It's my fault that I couldn't explain it properly in #502. 
What we actually need is a list of all validator addresses:
```go
	Validators() []crypto.Address
```

store/mock.go
```diff
@@ -15,18 +15,18 @@ import (
...
 
 type MockStore struct {
-	Blocks     map[uint32]block.Block
-	Accounts   map[crypto.Address]account.Account
-	Validators map[crypto.Address]validator.Validator
-	LastCert   *block.Certificate
-	LastHeight uint32
+	Blocks       map[uint32]block.Block
+	Accounts     map[crypto.Address]account.Account
+	ValidatorMap map[crypto.Address]validator.Validator
```
b00f, line 20: I understand your point about renaming Validators to ValidatorMap. However, to maintain consistency with the rest of the code, we have two options:

1- Don't rename it.
2- Rename all variables.

I prefer the first option because it keeps the PR atomic. Using Validators here is shorter and more suitable for the `Mock` object.

www/grpc/blockchain.go
```diff
@@ -218,8 +218,8 @@ func (s *blockchainServer) GetValidator(_ context.Context,
 
 func (s *blockchainServer) GetValidators(_ context.Context,
 	_ *pactus.GetValidatorsRequest) (*pactus.GetValidatorsResponse, error) {
-	validators := s.state.CommitteeValidators()
-	validatorsInfo := make([]*pactus.ValidatorInfo, 0)
```
b00f, line 222: We need to update proto to return a list of address string.

b00f (changes requested):

state/facade.go
Ja7ad: @b00f It's not `Validators()`, it's `ValidatorsAddress()`

store/mock.go
Ja7ad: It is not allowed to declare Method `Validators()` because we have a `Validators` field in struct and the compiler does not allow that.

b00f: Rename it to `ValidatorAddressess`

state/facade.go
b00f: `ValidatorAddresses()` sounds better. Thanks

b00f:
@Ja7ad Please rebase the PR. It has some conflicts.

store/validator.go
```diff
@@ -51,6 +51,14 @@ func (vs *validatorStore) hasValidator(addr crypto.Address) bool {
 	return ok
 }
 
+func (vs *validatorStore) ValidatorAddresses() []crypto.Address {
+	validatorAddresses := make([]crypto.Address, 0, len(vs.numberMap))
```
b00f, line 55: Here, better we use `addressMap`, It should be faster, since we don't need to calculate the address from the public key of the validator. The validator's address is the map key:

```go
for addr, _ := range vs.addresMap {
...
	}
```

Suggestion:
I recommend to rename `validatorAddresses` to `addrs`. This is a local variable within a short scope, a shorter name could be more readable.

store/mock.go
```diff
@@ -104,6 +104,13 @@ func (m *MockStore) HasValidator(addr crypto.Address) bool {
 	_, ok := m.Validators[addr]
 	return ok
 }
+func (m *MockStore) ValidatorAddresses() []crypto.Address {
+	validatorAddresses := make([]crypto.Address, 0, len(m.Validators))
+	for _, val := range m.Validators {
```
b00f, line 109: You can get the address from the map key.

www/grpc/blockchain.go
```diff
@@ -219,13 +219,23 @@ func (s *blockchainServer) GetValidator(_ context.Context,
 func (s *blockchainServer) GetValidators(_ context.Context,
```
b00f, line 219: I feel we don't need this function. This is replaced with `GetValidatorAddresses`. Isn't it?

Ja7ad: This method return list of commitee validators, if we don't need to this method we can remove in new issue.

b00f:
@Ja7ad I noticed a [failure](https://github.com/pactus-project/pactus/actions/runs/5297104187/jobs/9588654356#step:4:156) in `TestIterateValidators` tests in store module.
I guess the [random number](https://github.com/pactus-project/pactus/blob/5e2f1357a770a84454f2c5b6f8eb5e6f30518936/store/validator_test.go#LL162C5-L162C5) was zero and therefore the iterator didn't iterate. We can avoid this case by increasing the `total` by 1 : 
```go
	total := util.RandInt32(100) + 1
```

Similarly `TestAccountByAddress` [test](https://github.com/pactus-project/pactus/blob/main/store/account_test.go#L121) also should be updated. It is not related to this PR, but since it happens here for the first time and it is small matter, let's fix it here.

store/validator_test.go
```diff
@@ -66,6 +66,35 @@ func TestValidatorBatchSaving(t *testing.T) {
 	})
 }
 
+func TestValidatorAddresses(t *testing.T) {
```
b00f, line 69: Let's keep this test simpler:

```go
        total :=  util.RandInt32(100)
        addrs1 := make([]crypto.Address, total)
		for i := int32(0); i < total; i++ {
			val, _ := validator.GenerateTestValidator(int32(i))
			tStore.UpdateValidator(val)
			addrs1 = append(addrs1, val.Address())
		}
		
		addrs2 := store.ValidatorAddresses()
		assert.EqualsElements(t, addr1, addr2)
```

store/account_test.go
```diff
@@ -118,7 +118,7 @@ func TestAccountByNumber(t *testing.T) {
 func TestAccountByAddress(t *testing.T) {
 	setup(t)
 
-	total := util.RandInt32(100)
+	total := util.RandInt32(100) + 1
```
b00f, line 121: Opps, We should do it for `TestIterateAccounts` test. No har to have it here. Is there any other places?

Ja7ad: Yes


**store/validator_test.go**

```go
func TestAccountBatchSaving(t *testing.T) {
	setup(t)

	total := util.RandInt32(100)
	t.Run("Add some accounts", func(t *testing.T) {
		for i := int32(0); i < total; i++ {
			acc, signer := account.GenerateTestAccount(i)
			tStore.UpdateAccount(signer.Address(), acc)
		}
		assert.NoError(t, tStore.WriteBatch())
		assert.Equal(t, tStore.TotalAccounts(), total)
	})

	t.Run("Close and load db", func(t *testing.T) {
		tStore.Close()
		store, _ := NewStore(tStore.config, 21)
		assert.Equal(t, store.TotalAccounts(), total)
	})
}
```

Ja7ad: > Opps, We should do it for `TestIterateAccounts` test. No harm to have it here. Is there any other places?

If `total := 0` for-loop has been skip.

Ja7ad:
@b00f all thing changes done

b00f (approved):

b00f:
@Ja7ad Thanks

PR #518, opened 2023-06-18
refactor: optimization of total power calculation

b00f:
## Description

Prior to this PR, the total power (the total staked coins) was calculated at each height by iterating over all validators and summing up the power (stakes). This was not an efficient method for calculating total power.

With this PR, a more efficient approach has been implemented. Now, whenever a validator bonds or unbonds a stake, we record this as a power delta (change) and apply it to the total power. This removes the need for iterating over all validators at every height, thereby optimizing the total power calculation process.


Ja7ad (approved):

Ja7ad:
@b00f All thing is okay for merge

Issue #528, opened 2023-06-20
HTTP: Add Missing APIs for Account and Validator By Number


b00f:
## Describe the problem you'd like to have solved

The HTTP module in Pactus works like a simple HTTP server to provide some information about the node. 
You can think of it as a very basic block explorer.

Currently, there are two APIs missing in the HTTP module: `GetAccountByNumber` and `GetValidatorByNumber`.

### Describe the ideal solution

The first thing you need to do is to add these routes:

```
/account/number/{number}
/validator/number/{number}
```
Then add the `GetAccountByNumberHandler` handlers inside [account.go](https://github.com/pactus-project/pactus/blob/main/www/http/account.go) and  `GetValidatorByNumberHandler` inside [validator.go](https://github.com/pactus-project/pactus/blob/main/www/http/validator.go).

Please don't forget to write tests for both APIs.

kehiy:
hello.
I'll start working on this issue!

kehiy:
chek #541

kehiy:
fixed in: #549

kehiy:
@b00f should we close this issue?

b00f:
Yes, we can close it now. Thanks @kehiy

PR #549, opened 2023-06-27
fix(HTTP): Missing handlers

kehiy:
## Description
I worte getAccountByNumber and getValidatorByNumber handler and also test for the handlers.

## Related issue(s)
feat: #528


kehiy:
@b00f I think you can review this PR know!

www/http/account.go
```diff
@@ -26,3 +29,36 @@ func (s *Server) GetAccountHandler(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) GetAccountByNumberHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+
+	num, err := strconv.ParseInt(vars["number"], 10, 32)
+	if err != nil {
+		s.writeError(w, err)
+		return
+	}
+
+	if num > math.MaxInt32 || num < math.MinInt32 {
```
b00f, line 43: This check here is not necessary. HTTP is just a wrapper and should have not logic.

www/http/validator.go
```diff
@@ -30,3 +34,41 @@ func (s *Server) GetValidatorHandler(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) GetValidatorByNumberHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+
+	num, err := strconv.ParseInt(vars["number"], 10, 32)
+	if err != nil {
+		s.writeError(w, err)
+		return
+	}
+
+	if num > math.MaxInt32 || num < math.MinInt32 {
```
b00f, line 48: This check is not necessary. HTTP is just a wrapper and should have any logic. 
Please don't forget to remove the tests for this block

b00f (approved):

b00f:
@kehiy Everything looks good. Please not that HTTP is a just wrapper over gRPC. 
So, it is better to just call the gRPC, and let gRPC validates the data

kehiy:
> @kehiy Everything looks good. Please not that HTTP is a just wrapper over gRPC. So, it is better to just call the gRPC, and let gRPC validates the data

that't mean I need just remove the checks and test for them?

www/http/account_test.go
```diff
@@ -65,4 +66,43 @@ func TestAccount(t *testing.T) {
 		assert.Equal(t, w.Code, 400)
 		fmt.Println(w.Body)
 	})
+
+	t.Run("Shall return an account", func(t *testing.T) {
```
b00f, line 70: Let's move these tests to a new function: `TestAccountByNumber`

Also, try to cover the situation that account can't be found, like number is `acc.Number()+1`

www/http/validator_test.go
```diff
@@ -63,4 +64,45 @@ func TestValidator(t *testing.T) {
 		assert.Equal(t, w.Code, 400)
 		fmt.Println(w.Body)
 	})
+
+	t.Run("Shall return a validator", func(t *testing.T) {
```
b00f, line 68: Let's move these tests to a new function: `TestValidatorByNumber`

Also, try to cover the situation that validator can't be found, like number is `val.Number()+1`

b00f (approved):

b00f:
> > @kehiy Everything looks good. Please not that HTTP is a just wrapper over gRPC. So, it is better to just call the gRPC, and let gRPC validates the data
> 
> that't mean I need just remove the checks and test for them?

Correct. It is OK now

b00f (approved):

b00f (approved):

Issue #551, opened 2023-06-27
Removing OpenURLInBrowser test


b00f:
## Description

I noticed something strange today. When I ran the `make test`, my browser suddenly opened new tabs for example.com and pactus.com. 
After searching, I found out that the `TestOpenURLInBrowser` test was responsible for opening these new tabs. 
This is not a good test.  let's remove it and move the `OpenURLInBrowser` function from the `util` package to the `gui` package.

PR #553, opened 2023-06-28
fix: implemented restore wallet base on input seed

### LINKED ISSUE #522: GUI: Implementing restore wallet from seed phrase

Currently, the only way to initialize a node is by creating a new seed phrase (mnemonic). 
There is an option in the initialization assistant dialog for users to "Restore a wallet from the seed phrase", but this feature is currently inactive. We need to implement this feature.
Screenshot from 2023-06-20 00-28-26

The solution should be implemented within [startup_assistant](https://github.com/pactus-project/pactus/blob/main/cmd/gtk/startup_assistant.go#LL72C6-L72C14),  similar to `pageSeed`. We need to add a new page called `pageSeedRestore`. On this page, the user can enter enter the seed phrase. We can use the `bip39.IsMnemonicValid` API to verify the validity of the seed phrase. If the seed phrase is valid, the user can then proceed to the `passwordPage`.

In the assistant dialog, the sub-pages should change based on the user's choice on the first page (Initialize mode). If a user selects "Restore a wallet from the seed phrase", the pages should follow this sequence: Initialize Mode -> Restore from seed -> Password -> ... 
This process should be dynamic. If the user selects "Create a new wallet from scratch", the sequence should revert to the current sequence.

After the seed phrase has been entered, the wallet can be created (or recovered) on the final page.

Currently, there is no easy work-around for this problem. The only alternative is for users to manually use `pactus-wallet` to recover their wallet, and then replace it with the `default_wallet`.

Ja7ad:
## Description

This PR enables users to restore the `default_wallet` in the GUI within the initialize assistant dialog. 
To ensure proper page navigation, it jumps between pages based on the initialize mode.

## Related issues

Fixed #522 and #551

### DIFF
--- a/cmd/gtk/main_window.go
+++ b/cmd/gtk/main_window.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/gotk3/gotk3/gtk"
 	"github.com/pactus-project/pactus/crypto"
-	"github.com/pactus-project/pactus/util"
 )
 
 //go:embed assets/ui/main_window.ui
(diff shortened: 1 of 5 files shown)


b00f (approved):

b00f:
@Ja7ad Thanks

Issue #147, opened 2021-04-25
When the last time a peer was online?


b00f:
We have the [list of connected peers](https://github.com/zarbchain/zarb-go/blob/main/sync/peerset/peer_set.go). 
We should be able to know when was the last time a peer was online. This will help to manage the peer list and remove offline peers from the list

b00f:
It has fixed here: https://github.com/pactus-project/pactus/commit/f5c78e6fba96c788ba8f7bab76e102ee38408637

PR #555, opened 2023-06-30
feat: last seen of a peer

amirvalhalla:
## Description

the last seen peer to detecting is offline or not

## Related issue(s)

fixed https://github.com/pactus-project/pactus/issues/147

### DIFF
--- a/sync/peerset/peer.go
+++ b/sync/peerset/peer.go
@@ -49,3 +49,7 @@ func (p *Peer) IsBanned() bool {
 func (p *Peer) IsNodeNetwork() bool {
 	return util.IsFlagSet(p.Flags, PeerFlagNodeNetwork)
 }
+
+func (p *Peer) LastSeenAt() time.Time {
+	return p.LastSeen
+}


sync/peerset/peer.go
```diff
@@ -49,3 +49,7 @@ func (p *Peer) IsBanned() bool {
 func (p *Peer) IsNodeNetwork() bool {
 	return util.IsFlagSet(p.Flags, PeerFlagNodeNetwork)
 }
+
+func (p *Peer) LastSeenAt() time.Time {
```
b00f, line 53: I don't think we need this function. 

The `LastSeen` property is exported and it is reported in gRPC [here](https://github.com/pactus-project/pactus/blob/f01370fe659c6f9e42cc2f116866562d22561501/www/grpc/network.go#L43)

amirvalhalla: so I think I didn't understand well the purpose of this issue , what feature exactly should implement?

b00f: It was bad of me. It was fixed here: https://github.com/pactus-project/pactus/commit/f5c78e6fba96c788ba8f7bab76e102ee38408637

PR #560, opened 2023-06-30
style: rename ParsMessage to ParseMessage

amirvalhalla:
## Description

there was a typo about ParsMessage in messageHandler interface which I've fixed and converted to ParseMessage


b00f (approved):

b00f:
@amirvalhalla Thanks

Issue #368, opened 2022-09-03
Set Context properly


b00f:
Some modules have their own Context. It's better to define a global context in the start function at[ node.go](https://github.com/zarbchain/zarb-go/blob/main/node/node.go) and pass it to other modules.
By stopping the node, the context can be closed for all modules.

PR #561, opened 2023-07-01
feat: GarbageCollector

### LINKED ISSUE #546: Garbage collection for peerset

Once a node joins the network, it sends `hello` [message](https://github.com/pactus-project/pactus/blob/a75fe24372bd3cd35d6a6c0d725f26e84a1c0357/sync/bundle/message/hello.go#L20) and announce its peer ID and consensus keys. 
However, we currently don't have a mechanism to remove a node from the [peer list](https://github.com/pactus-project/pactus/blob/a75fe24372bd3cd35d6a6c0d725f26e84a1c0357/sync/peerset/peer_set.go#L19C23-L19C24) once it goes offline.

We need to establish a mechanism for garbage collection. A garbage node should have a clear definition.
A node can be considered "garbage" if we haven't heard from it in the last few hours.
Each peer has a property named [last_seen](https://github.com/pactus-project/pactus/blob/a75fe24372bd3cd35d6a6c0d725f26e84a1c0357/sync/peerset/peer.go#L24) that gets updated each time we receive data from that node.
We can use this property to calculate how long a node has been offline. 
If a node has been offline for more than a day, we should remove it from the peer list. 
The garbage collector check could be scheduled to run daily

Currently, the only workaround is to restart the node.

kehiy:
## Description
I added the Garbage Collector function for removing the offline peers each 24 hours

## Related issue

- Fixes #546

### DIFF
--- a/sync/peerset/peer_set.go
+++ b/sync/peerset/peer_set.go
@@ -1,6 +1,7 @@
 package peerset
 
 import (
+	"context"
 	"sync"
 	"time"
 
@@ -8,10 +9,6 @@ import (
 	"github.com/pactus-project/pactus/crypto/bls"
 	"github.com/pactus-project/pactus/util"
 )
-
-// TODO:
-// - Add tests for peerset
-// - Implementing garbage collection for peerset
 
 type PeerSet struct {
 	lk sync.RWMutex
@@ -347,4 +344,24 @@ func (ps *PeerSet) StartedAt() time.Time {
 	defer ps.lk.RUnlock()
 
 	return ps.startedAt
+}
+
+func (ps *PeerSet) GarbageCollector(ctx context.Context, timeout time.Duration) {
+	for {
+		timer := time.NewTimer(timeout)
+		select {
+		case <-timer.C:
+			for _, peer := range ps.peers {
+				now := time.Now()
+				diff := now.Sub(peer.LastSeen)
+
+				if diff.Hours() >= 24 {
+					ps.RemovePeer(peer.PeerID)
+				}
+			}
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		}
+	}
 }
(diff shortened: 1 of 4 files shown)


kehiy:
@b00f you can review it know
I understood the test completely

b00f:
@kehiy update [RandomString](https://github.com/pactus-project/pactus/blob/f01370fe659c6f9e42cc2f116866562d22561501/util/testsuite/testsuite.go#L140) function and change 52 to 51, will fix the broken test. We have just recently introduced the TestSuite.

b00f:
We need to make sure that Context will be set properly #368

kehiy:
@b00f I changed it to 51 and first issue is fixed!
but there is another issue:
```
Error Trace:	/home/kehiy/code/pactus/sync/peerset/peer_set_test.go:296
            	Error:      	Should be zero, but was 2
            	Test:       	TestGarbageCollector/Removing_peers
```
I think the problem is from this lines:
```go
time.Sleep(1 * time.Second)
		cancel()
```

kehiy:
> We need to make sure that Context will be set properly #368

Isn't it better to solve them all at once in a single pull request?

b00f:
Give me the seed, then I can reproduce it here

b00f:
> Isn't it better to solve them all at once in a single pull request?

Yes. Opening new PR for this small fixes is not a good option. 

Here, I think we can fix it by reducing `clearanceTimeout` to 500 millisecond

kehiy:
> > Isn't it better to solve them all at once in a single pull request?
> 
> Yes. Opening new PR for this small fixes is not a good option.
> 
> Here, I think we can fix it by reducing `clearanceTimeout` to 300 millisecond

i fixed it

yes, maybe i go to work on this after resolving the address validator

kehiy:
@b00f its still need approving

b00f:
@kehiy please rebase your branch. It has some conflict with the codebase.

kehiy:
> @kehiy please rebase your branch. It has some conflict with the codebase.

can I resolve conflicts here? in github??
I think its mush easier

PR #562, opened 2023-07-01
feat: Get validator info from wallet

jwmdev:
## Description

Getting validator info from wallet given the validator address

## Related issue(s)

No issue

### DIFF
--- a/sync/bundle/message/hello.go
+++ b/sync/bundle/message/hello.go
@@ -18,15 +18,15 @@ const (
 )
 
 type HelloMessage struct {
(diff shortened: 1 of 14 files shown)


b00f (approved):

b00f:
@jwmdev

There's no harm with returning `validatorInfo` in the wallet. However, I have a few questions:

1- Until now, we've added some APIs in the wallet to retrieve **partial** information about the validator, such as Stake or Sequence. This API return all information about the validator. Which one should we maintain, the partial information or the full information?

2- We should also implement the same for the account.

Issue #524, opened 2023-06-20
Address validation for the network module


b00f:
## Description

We currently do not validate the [network addresses](https://docs.libp2p.io/concepts/fundamentals/addressing/) in the config file. So, if a user enters an incorrect address, the node won't start (an incorrect listen address can even cause a system crash). We need to ensure all addresses, both relay and listen addresses, are valid in the [SanityCheck](https://github.com/pactus-project/pactus/blob/01d155b09c7db3ab279668f7973f7ebc9194cc89/network/config.go#L65) function. 

### How To Reproduce

1. Open the config file.
2. Change the listen addresses and relay addresses to an incorrect string.

### What Happened

The node didn't start. It may also cause a system crash.

### Expected Behavior

Should return invalid address error before starting the node.
Write tests for the invalid addressess.

kehiy:
check: #565

kehiy:
@b00f we can close it now!

PR #565, opened 2023-07-01
feat: Check node address is valid

kehiy:
## Description

This PR checks the validity of addresses inside the network's  [config.go](https://github.com/pactus-project/pactus/blob/01d155b09c7db3ab279668f7973f7ebc9194cc89/network/config.go#L65) and returns error if any of the addresses is invalid. 

## Related issue
Fixing #524


network/config.go
```diff
@@ -61,12 +62,34 @@ func DefaultConfig() *Config {
...
+	for _, addr := range address {
+		isValid := isValidAddress(addr)
+		if !isValid {
+			return false
+		}
+	}
+	return true
+}
+
+func isValidAddress(address string) bool {
```
b00f, line 75: Why do we return `bool` and not `error`? An error message would give users some information about why the address is invalid.
This function can be merged with `validateAddresses`.

network/config.go
```diff
@@ -61,12 +62,34 @@ func DefaultConfig() *Config {
 	}
 }
 
+func validateAddress(address []string) bool {
```
b00f, line 65: `validateAddress` should be `validateAddresses` and the function should return `error`.

network/config_test.go
```diff
@@ -15,3 +15,19 @@ func TestConf(t *testing.T) {
 	conf.EnableRelay = false
 	assert.NoError(t, conf.SanityCheck())
 }
+
+func TestValidateAddress(t *testing.T) {
+	validAddresses := []string{
+		"/ip4/127.0.0.1/tcp/21777/p2p/QmUv7r8T5NT3xQ2Zzvbg53uSdVccQ6LbdY2wAju14HsWyc",
```
b00f, line 21: For valid address you can use: "/ip4/127.0.0.1"
For invalid addresses you can use: "", "/ip4", "127.0.0.1"

Keep test simple

b00f (approved):

network/config_test.go
```diff
@@ -15,3 +15,19 @@ func TestConf(t *testing.T) {
 	conf.EnableRelay = false
 	assert.NoError(t, conf.SanityCheck())
 }
+
+func TestValidateAddress(t *testing.T) {
```
b00f, line 19: It's not recommended to test unreported functions. 
So, move these tests into  `TestDefaultConfigCheck` (rename `TestConf` to `TestDefaultConfigCheck`.

b00f (approved):

kehiy:
@b00f this is better version 
i think the print statement in validate function is still unnecessary, is that?

network/config.go
```diff
@@ -61,12 +62,25 @@ func DefaultConfig() *Config {
 	}
 }
 
+func validateAddresses(address []string) error {
+	for _, addr := range address {
+		_, err := multiaddr.NewMultiaddr(addr)
+		if err != nil {
+			fmt.Println("Failed to parse address: ", err)
```
b00f, line 69: No need to print log here.

network/config.go
```diff
@@ -61,12 +62,25 @@ func DefaultConfig() *Config {
...
+			return errors.Errorf(errors.ErrInvalidAddress, "invalid relay and listen address")
```
b00f, line 70: Just `return err`

network/config.go
```diff
@@ -61,12 +62,25 @@ func DefaultConfig() *Config {
...
+}
+
 // SanityCheck performs basic checks on the configuration.
 func (conf *Config) SanityCheck() error {
 	if conf.EnableRelay {
 		if len(conf.RelayAddrs) == 0 {
 			return errors.Errorf(errors.ErrInvalidConfig, "at least one relay address should be defined")
 		}
 	}
+	validateAddresses(conf.Listens)
```
b00f, line 83:
```
if err := validateAddresses(conf.Listens); err != nil {
    return err
}
```

b00f (changes requested):

network/config.go
kehiy: should I add this for `conf.RelayAddrs` too?
```go
if err := validateAddresses(conf.Listens); err != nil {
    return err
}
if err := validateAddresses(conf.RelayAddrs); err != nil {
    return err
}
```
like this

network/config.go
```diff
@@ -61,12 +62,28 @@ func DefaultConfig() *Config {
...
 func (conf *Config) SanityCheck() error {
 	if conf.EnableRelay {
 		if len(conf.RelayAddrs) == 0 {
 			return errors.Errorf(errors.ErrInvalidConfig, "at least one relay address should be defined")
 		}
 	}
+	if err := validateAddresses(conf.Listens); err != nil {
+		return err
+	}
+	if err := validateAddresses(conf.RelayAddrs); err != nil {
```
b00f, line 85: You can fix the linting issue by: `return validateAddresses(conf.RelayAddrs)`

kehiy: I get an unreachable warning in this case:
```go
// SanityCheck performs basic checks on the configuration.
func (conf *Config) SanityCheck() error {
	if conf.EnableRelay {
		if len(conf.RelayAddrs) == 0 {
			return errors.Errorf(errors.ErrInvalidConfig, "at least one relay address should be defined")
		}
	}
	return validateAddresses(conf.RelayAddrs)
	return validateAddresses(conf.Listens)
}
```

b00f: Only the last `return validateAddresses(conf.Listens)`
You know it nags that why you created new variable. Just return from it. So it will be like:
```go
if err:= validateAddresses(conf.RelayAddrs); err != nil {
   return err
}
return validateAddresses(conf.Listens) 
```

kehiy: yeah, I got it! that was fixed.

b00f (approved):

network/config_test.go
```diff
@@ -6,12 +6,25 @@ import (
...
 	conf.EnableRelay = false
 	assert.NoError(t, conf.SanityCheck())
+
+	inValidAddresses := []string{
+		"/ip4/1wg.1/tcp/21777/p2p8VccQ8888888Wyc",
+		"",
+		"/ip4",
+		"127.0.0.1",
+	}
+	assert.Error(t, validateAddresses(inValidAddresses))
```
b00f, line 24:
```
conf.Listens = []string{""}
assert.Error(t, conf.SanityCheck())

conf.Listens = []string{"127.0.0.1"}
assert.Error(t, conf.SanityCheck())

conf.Listens = []string{"/ip4"}
assert.Error(t, conf.SanityCheck())

conf.Listens = []string{"/ip4/127.0.0.1"}
assert.NoError(t, conf.SanityCheck())

conf.RelayAddresses = []string{"/ip4"}
assert.Error(t, conf.SanityCheck())

conf.RelayAddresses = []string{"/ip4/127.0.0.1"}
assert.NoError(t, conf.SanityCheck())
```

b00f (approved):

b00f:
Awesome @kehiy

PR #570, opened 2023-07-04
feat: License year update

kehiy:
## Description
just update 2020 to 2023.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Pactus blockchain
+Copyright (c) 2023 Pactus blockchain
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


amirvalhalla (approved):
@kehiy  thanks

b00f:
@kehiy @amirvalhalla 

I am going to revert this change. Because, I checked some MIT projects like [NodeJS](https://github.com/nodejs/node/blob/main/LICENSE), [Angular](https://github.com/angular/angular.js/blob/master/LICENSE), [curl](https://github.com/curl/curl/blob/master/COPYING). Also asked ChatGPT:

> In MIT license, what should be the year?
> 
> In the MIT License, the year should be the year that the work was published or created. The license is typically included when the software is published, so the year would be when the software was first made available to the public.

What we know that the year should be the year that the project is [published](https://github.com/pactus-project/pactus/commit/bb907cde5ec313300bf3ab73e7f08a0ddce50b96). Clearly here is 2020.

Some projects follow this pattern: `<Publish year> - <This year>`: like `2000 - 2023`. I found this post on [curl](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/) project. However, following this style would also require us to update the license every year, which could be inconvenient.

amirvalhalla:
@b00f thanks.

kehiy:
> @kehiy @amirvalhalla
> 
> I am going to revert this change. Because, I checked some MIT projects like [NodeJS](https://github.com/nodejs/node/blob/main/LICENSE), [Angular](https://github.com/angular/angular.js/blob/master/LICENSE), [curl](https://github.com/curl/curl/blob/master/COPYING). Also asked ChatGPT:
> 
> > In MIT license, what should be the year?
> > In the MIT License, the year should be the year that the work was published or created. The license is typically included when the software is published, so the year would be when the software was first made available to the public.
> 
> What we know that the year should be the year that the project is [published](https://github.com/pactus-project/pactus/commit/bb907cde5ec313300bf3ab73e7f08a0ddce50b96). Clearly here is 2020.
> 
> Some projects follow this pattern: `<Publish year> - <This year>`: like `2000 - 2023`. I found this post on [curl](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/) project. However, following this style would also require us to update the license every year, which could be inconvenient.

sure, we should follow the best way.

PR #574, opened 2023-07-05
feat!: adding MinimumStake parameter

### LINKED ISSUE #501: Adding MinimumStake to the consensus parameters

Adding `MinimumStake` to the consensus parameters prevents creating dummy validators, specially in the Mainnet.
Spammers might create thousands of validators with small stakes. This has no effect on the consensus mechanism or the sortition algorithm.

Adding a new field in the [consensus parameters](https://github.com/pactus-project/pactus/blob/main/types/param/param.go) for `MinimumStake`. The default value can set to 1 PAC coin.

kehiy:
## Description

This PR adds the `MinimumStake` parameter to the [consensus parameters](https://github.com/pactus-project/pactus/blob/236b5cac9e124a3baea8162478ae01c958e2f607/types/param/param.go#L5) and the default value is set to 1 PAC coin (10^9).
Consequently, all bond transaction will reject if their value is less than the `MinimumStake`.

## Related issue(s)

- Fixes #501

BREAKING CHANGE


Ja7ad:
> ## Description
> I added `MinimumStake` amount as 1 PAC coin.
> there is just one issue with the test and it is:
> the random amount and fee always going to be less than `MinimumStake` we should change the random fee and amount or maybe the test.
> 
> ## Related issue(s)
> - Fixes #501

Please send your PR on **mainnet** branch.

kehiy:
> > ## Description
> > I added `MinimumStake` amount as 1 PAC coin.
> > there is just one issue with the test and it is:
> > the random amount and fee always going to be less than `MinimumStake` we should change the random fee and amount or maybe the test.
> > 
> > ## Related issue(s)
> > - Fixes #501
> 
> Please send your PR on **mainnet** branch.

@kehiy sure!
I will send this PR and the GC pr to mainnet branch, but I should notice that this PR need a small change in merge time....

types/param/param.go
```diff
@@ -14,6 +14,7 @@ type Params struct {
 	FeeFraction               float64 `cbor:"9,keyasint"`
 	MinimumFee                int64   `cbor:"10,keyasint"`
 	MaximumFee                int64   `cbor:"11,keyasint"`
+	MinimumStake              int64   `cobr:"12,keyasint"`
 	MaximumStake              int64   `cbor:"12,keyasint"`
```
b00f, line 18:
```suggestion
	MaximumStake              int64   `cbor:"13,keyasint"`
```

kehiy:
@b00f I commited yout sugget, but don't forget to solve test issue in mergeing time.

b00f:
@kehiy 

I think we can fix test with small modification for `randomAmountAndFee`, like:

```go
func (td *testData) randomAmountAndFee(min int64, max int64) (int64, int64) {
	amt := td.RandInt64NonZero(max)
	for amt < min {
		amt = td.RandInt64NonZero(max)
	}

	fee := int64(float64(amt) * td.sandbox.Params().FeeFraction)
	return amt, fee
}
```

kehiy:
@b00f 
I test it, but that doesn't work!

kehiy:
@b00f some other tests going to be filled, can you review them?

execution/executor/transfer_test.go
```diff
@@ -57,7 +61,7 @@ func TestExecuteTransferTx(t *testing.T) {
 	senderAddr, senderAcc := td.sandbox.TestStore.RandomTestAcc()
 	senderBalance := senderAcc.Balance()
 	receiverAddr := td.RandomAddress()
-	amt, fee := td.randomAmountAndFee(senderBalance)
+	amt, fee := td.randomAmountAndFee(td.sandbox.TestParams.MinimumFee, senderBalance)
```
b00f, line 64: Only set minimum for the Bond transactions. for other transaction keep it zero.

b00f:
@kehiy Please use `make test` to check it locally. I can see some issues in tests.
Also, for non-bond transactions, set minimum to zero.

Thanks buddy

kehiy:
@b00f I fixed it, the last issue is `pool_test.go`

execution/executor/transfer_test.go
```diff
@@ -44,10 +44,14 @@ func (td *testData) checkTotalCoin(t *testing.T, fee int64) {
...
-	amt := td.RandInt64(max / 2)
+func (td *testData) randomAmountAndFee(min int64, max int64) (int64, int64) {
+	amt := td.RandInt64NonZero(max)
+	for amt < min {
+		amt = td.RandInt64NonZero(max)
+	}
+
 	fee := int64(float64(amt) * td.sandbox.Params().FeeFraction)
-	return amt, fee
+	return fee, amt
```
b00f, line 54: why??

execution/executor/bond_test.go
```diff
@@ -80,6 +79,14 @@ func TestExecuteBondTx(t *testing.T) {
 		assert.Equal(t, errors.Code(err), errors.ErrInvalidPublicKey)
 	})
 
+	t.Run("Should fail, amount less than MinimumStake", func(t *testing.T) {
+		trx := tx.NewBondTx(td.stamp500000, senderAcc.Sequence()+1, senderAddr,
```
b00f, line 83: I think we changed `stamp500000` to `randStamp`.

@kehiy run `make unit_test` locally.

execution/executor/transfer_test.go
kehiy: that was return them reverse I think
but I can change order of them when call it

execution/executor/bond_test.go
kehiy: the errors is from state test

execution/executor/bond_test.go
```diff
@@ -116,7 +123,7 @@ func TestBondInsideCommittee(t *testing.T) {
 	exe2 := NewBondExecutor(false)
 	senderAddr, senderAcc := td.sandbox.TestStore.RandomTestAcc()
 	senderBalance := senderAcc.Balance()
-	fee, amt := td.randomAmountAndFee(senderBalance)
+	amt, fee := td.randomAmountAndFee(0, senderBalance)
```
b00f, line 126: Thanks for fixing this mistake.

execution/executor/transfer.go
```diff
@@ -54,6 +54,9 @@ func (e *TransferExecutor) Execute(trx *tx.Tx, sb sandbox.Sandbox) error {
 		return errors.Errorf(errors.ErrInvalidSequence,
 			"expected: %v, got: %v", senderAcc.Sequence()+1, trx.Sequence())
 	}
+	if senderAcc.Balance() < pld.Amount+trx.Fee() {
```
b00f, line 57: There is same logic several line above, Remove that.
The point is, first check sequence then balance.

genesis/genesis_test.go
```diff
@@ -57,8 +58,8 @@ func TestGenesisTestNet(t *testing.T) {
 	genTime, _ := time.Parse("2006-01-02", "2023-05-08")
 	assert.Equal(t, g.GenesisTime(), genTime)
 	assert.Equal(t, g.Params().BondInterval, uint32(120))
-
-	expected, _ := hash.FromString("d5b41d8c2d45a951329512915ae7b2f4a445aeda380a1d40c50b5a5743dfb3fa")
+	fmt.Println(g.Hash())
```
b00f, line 61: Remove this line, please

b00f (changes requested):

genesis/genesis_test.go
kehiy: fixed

execution/executor/transfer.go
kehiy: fixed

themantre (approved):

kehiy:
@b00f checks are passed!

PR #576, opened 2023-07-05
refactor: public key aggregate

RiXelanya:
## Description

Extract Public Key aggregate function from the VerifyAggregated function. Write some tests. There is still a problem

```
func TestPublicKeyAggregate(t *testing.T) {
	pub1, _ := bls.PublicKeyFromString(
		public key string 1)
	pub2, _ := bls.PublicKeyFromString(
		public key string 2)
	agg, _ := bls.PublicKeyFromString(
		aggregated public key)
	assert.True(t, bls.PublicKeyAggregate([]*bls.PublicKey{pub1, pub2}).EqualsTo(agg))
	assert.False(t, pub1.EqualsTo(pub2))
}
```

I dont know how to generate the public key and aggregated public key string value to be used as examples in this test

## Related issue(s)

### DIFF
--- a/consensus/consensus_test.go
+++ b/consensus/consensus_test.go
@@ -290,7 +290,7 @@ func (td *testData) commitBlockForAllStates(t *testing.T) (*block.Block, *block.
 	sig2 := td.signers[1].SignData(sb).(*bls.Signature)
 	sig4 := td.signers[3].SignData(sb).(*bls.Signature)
 
-	sig := bls.Aggregate([]*bls.Signature{sig1, sig2, sig4})
+	sig := bls.SignatureAggregate([]*bls.Signature{sig1, sig2, sig4})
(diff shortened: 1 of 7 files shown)


b00f:
@RiXelanya Thanks. I just changed "refactor" to "fix" for a more accurate change log.

RiXelanya:
Do you have any recommendation for tools to generate BLS Key string? I still have a problem in writing test as I have described in the initial post. I still dont know tools to generate BLS public key string and the G2 addition

b00f:
> Do you have any recommendation for tools to generate BLS Key string?

You can simple use `fmt.Println(pub.String())`. I have plans to revise these tests again because they are very important. Specifically, we need to write test for the [infinity](https://github.com/pactus-project/pactus/blob/81ab46e17b00f00c9f4d6c0f0ecc779d4b7d858f/crypto/bls/public_key_test.go#L127) public key

PR #579, opened 2023-07-06
fix: broadcast message with aggregated key

### LINKED ISSUE #547: Hello message should contains all public keys
## Description

Currently, the hello message only contains one public key (or consensus key). If a node has more than one consensus key, it broadcasts one hello message per key.

### How To Reproduce

We should revise the hello message to include all the consensus keys. To sign the message, we could aggregate all the keys and sign the message using the aggregated key.

~~This change needs to be **backward** compatible in Testnet.~~
~~We have two choices:~~
~~1. Define a new hello message (v2)~~
~~2. Modify the current hello message to support both versions~~

### What Happened

We receive multiple hello messages when a node joins the network.

### Expected Behavior

We should receive only one message per node.

RiXelanya:
## Description

I tried to make a new broadcast which includes the aggregate signature and public keys

## Related issue(s)

- Fixes #547

### DIFF
--- a/sync/bundle/message/hello.go
+++ b/sync/bundle/message/hello.go
@@ -18,15 +18,15 @@ const (
 )
 
 type HelloMessage struct {
-	PeerID      peer.ID        `cbor:"1,keyasint"`
-	Agent       string         `cbor:"2,keyasint"`
-	Moniker     string         `cbor:"3,keyasint"`
-	PublicKey   *bls.PublicKey `cbor:"4,keyasint,omitempty"`
+	PeerID      peer.ID          `cbor:"1,keyasint"`
+	Agent       string           `cbor:"2,keyasint"`
+	Moniker     string           `cbor:"3,keyasint"`
+	PublicKey   *bls.PublicKey   `cbor:"4,keyasint,omitempty"`
 	PublicKeys  []*bls.PublicKey `cbor:"9,keyasint,omitempty"`
-	Signature   *bls.Signature `cbor:"5,keyasint"`
-	Height      uint32         `cbor:"6,keyasint"`
-	Flags       int            `cbor:"7,keyasint"`
-	GenesisHash hash.Hash      `cbor:"8,keyasint"`
+	Signature   *bls.Signature   `cbor:"5,keyasint"`
+	Height      uint32           `cbor:"6,keyasint"`
+	Flags       int              `cbor:"7,keyasint"`
+	GenesisHash hash.Hash        `cbor:"8,keyasint"`
 }
 
 func NewHelloMessage(pid peer.ID, moniker string,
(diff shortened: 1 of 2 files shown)


amirvalhalla:
@RiXelanya 
please follow the PR title rules. read more:
[Contributing guidelines](https://github.com/pactus-project/pactus/blob/main/CONTRIBUTING.md)

b00f:
@RiXelanya 

From your PR, I can see that you got a little bit confused. I believe the main reason for this confusion is the need to maintain backward compatibility. It's okay, no problem. I'll try to explain it a little more here.

## sayHelloLagacy method

I suggest renaming the `sayHello` function to `sayHelloLegacy` and keeping it as it is. Please make sure to apply this change to the tests as well, so that they remain functional.

The `sayHelloLegacy` method helps us maintain backward compatibility, but we don't need to keep it forever. In a future release, once we confirm that the majority of the nodes are supporting the new format, we can safely delete it.

Adding comments to the code would be a good idea to provide clarity and context.
 
## sayHello method

Now we can implement `sayHello`, In this method first we can get the list of all public keys from the `sync.signers` and at the same time we calculate the aggregated signature.

Probably should be something like this:

```go
pubs := []*bls.PublicKey{}
sigs := []*bls.Signature{}

for _, signer := range sync.signers {
	signer.SignMsg(msg)
	pubs = append(pubkeys, msg.PublicKey)
	sigs = append(signs, msg.Signature)
}
aggSig := bls.SignatureAggregate(signs)
```

Now, we have the aggregated signature (only one signature) and list of all public keys.
Let's create another hello message and broadcast it. 

```go
msg.Signature = aggSig
msg.PublicKeys = pubs
msg.PublicKey = nil
```

## Signature Verification 

Now, let's move to the verification part. It is done inside the [SanityCheck](https://github.com/pactus-project/pactus/blob/6a16d7d3c8e27153baf353e366b3cecfa9af7200/sync/bundle/message/hello.go#L43C1-L43C1) function.
Here, we need to support both Legacy and new format.

How to figure out? It is easy. If `PulicKey` set, it is legacy version, it `PubliKeys` set it is new version.
So we will have something like:

```go
func (m *HelloMessage) SanityCheck() error {
	if m.Signature == nil {
		return errors.Error(errors.ErrInvalidSignature)
	}
	if m.PublicKey == nil &&  m.PublicKeys == nil {
		return errors.Error(errors.ErrInvalidPublicKey)
	}
	if m.PublicKey != nil {
             /// Legacy version
             return m.PublicKey.Verify(m.SignBytes(), m.Signature)
	} else {
	/// 1- Calculate the Aggregated Public key
	/// 2- Verify the signature
        }
}
```

RiXelanya:
Understood. I will start fixing it now. apologies for the late response. I was busy yesterday

themantre:
Fixed at #640

PR #582, opened 2023-07-08
fix: Set Context properly 

### LINKED ISSUE #368: Set Context properly

kehiy:
## Description

This PR introduces a global context and passed it to all modules when the node starts. By shutting down the node, the context will be cancelled and all modules will receive the cancellation signal. 

## Related issue(s)

- Fixes #368

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -26,7 +26,6 @@ type network struct {
 	// We should remove it from here and pass it as first argument of functions
 	// Adding these linter later:  contextcheck and containedctx
 	ctx            context.Context
-	cancel         func()
 	config         *Config
 	host           lp2phost.Host
 	mdns           *mdnsService
@@ -71,11 +70,11 @@ func loadOrCreateKey(path string) (lp2pcrypto.PrivKey, error) {
 	return key, nil
 }
 
-func NewNetwork(ctx context.Context, cancel func(), conf *Config) (Network, error) {
-	return newNetwork(ctx, cancel, conf, []lp2p.Option{})
+func NewNetwork(ctx context.Context, conf *Config) (Network, error) {
+	return newNetwork(ctx, conf, []lp2p.Option{})
 }
 
-func newNetwork(ctx context.Context, cancel func(), conf *Config, opts []lp2p.Option) (*network, error) {
+func newNetwork(ctx context.Context, conf *Config, opts []lp2p.Option) (*network, error) {
 	networkKey, err := loadOrCreateKey(conf.NetworkKey)
 	if err != nil {
 		return nil, errors.Errorf(errors.ErrNetwork, err.Error())
@@ -125,7 +124,6 @@ func newNetwork(ctx context.Context, cancel func(), conf *Config, opts []lp2p.Op
 
 	n := &network{
 		ctx:          ctx,
-		cancel:       cancel,
 		config:       conf,
 		host:         host,
 		eventChannel: make(chan Event, 100),
@@ -176,8 +174,6 @@ func (n *network) Start() error {
 }
 
 func (n *network) Stop() {
-	n.cancel()
-
 	if n.mdns != nil {
 		n.mdns.Stop()
 	}
--- a/network/network.go
+++ b/network/network.go
@@ -126,7 +126,7 @@ func newNetwork(conf *Config, opts []lp2p.Option) (*network, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	n := &network{
-		ctx:          ctx,
+		ctx:          ctx, //
 		cancel:       cancel,
 		config:       conf,
 		host:         host,
--- a/network/network.go
+++ b/network/network.go
@@ -71,11 +71,11 @@ func loadOrCreateKey(path string) (lp2pcrypto.PrivKey, error) {
 	return key, nil
 }
 
-func NewNetwork(conf *Config, ctx context.Context, cancel func()) (Network, error) {
-	return newNetwork(conf, []lp2p.Option{}, ctx, cancel)
+func NewNetwork(ctx context.Context, cancel func(), conf *Config) (Network, error) {
+	return newNetwork(ctx, cancel, conf, []lp2p.Option{})
 }
 
-func newNetwork(conf *Config, opts []lp2p.Option, ctx context.Context, cancel func()) (*network, error) {
+func newNetwork(ctx context.Context, cancel func(), conf *Config, opts []lp2p.Option) (*network, error) {
 	networkKey, err := loadOrCreateKey(conf.NetworkKey)
 	if err != nil {
 		return nil, errors.Errorf(errors.ErrNetwork, err.Error())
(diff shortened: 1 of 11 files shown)


amirvalhalla:
@kehiy 

I thinks there are 2 issue which you should fix in your code

1.  in `network/network.go` when you call `newNetwork` you passed `ctx`,`cancel` which their correct but after you created the `network` struct you used `n.ctx` in `newDHTService(n.ctx, n.host, kadProtocolID, conf.Bootstrap, n.logger)` and for 2 other sub methods just used `ctx` that all of them have same memory address for ctx but different style code. please use `n.ctx` for all of them to have a same style.

2. in `sync/sync.go` you used `ctx:=context.Background()` which I think it would be better to create a context with cancellation, so by that when you stop `sync` package it will call `cancel()` and `Done()` channel  of `ctx` will raise.


@b00f what is your opinion?

network/network.go
```diff
@@ -71,11 +71,11 @@ func loadOrCreateKey(path string) (lp2pcrypto.PrivKey, error) {
 	return key, nil
 }
 
-func NewNetwork(conf *Config) (Network, error) {
-	return newNetwork(conf, []lp2p.Option{})
+func NewNetwork(ctx context.Context, cancel func(), conf *Config) (Network, error) {
```
b00f, line 74: You should not pass cancel to the Network module.

node/node.go
```diff
@@ -46,7 +47,9 @@ func NewNode(genDoc *genesis.Genesis, conf *config.Config,
 		"version", version.Version(),
 		"network", genDoc.ChainType())
 
-	network, err := network.NewNetwork(conf.Network)
+	ctx, cancel := context.WithCancel(context.Background())
```
b00f, line 50: I think we need to add cancel to the node structure and call it when the node is shutting down

b00f (approved):

kehiy:
> @kehiy
> 
> I thinks there are 2 issues which you should fix in your code
> 
> 1. in `network/network.go` when you call `newNetwork` you passed `ctx`,`cancel` which it's correct but after you created the `network` struct you used `n.ctx` in `newDHTService(n.ctx, n.host, kadProtocolID, conf.Bootstrap, n.logger)` and for 2 other sub methods just used `ctx` that all of them have same memory address for ctx but different style code. please use `n.ctx` for all of them to have a same style.
> 2. in `sync/sync.go` you used `ctx:=context.Background()` which I think it would be better to create a context with cancellation, therefore when you stop `sync` package it will call `cancel()` and `Done()` channel  of `ctx` will raise.
> 
> @b00f what is your opinion?

sure, I think its true
probably i will fix it

b00f:
@Ja7ad please review this PR

network/network.go
kehiy: but, the network module needs the cancel!
should I remove it from the Network module completely?

node/node.go
kehiy: yes, I can do it, but just tell me what should I do about the network module cancel?

network/network.go
b00f: Why network module needs cancel?
This context is global and should be created when the node starts and cancelled when the node finishes.

kehiy: so, that's means I should remove the network cancel and add it to the node struct and call it when the node stopped.
right?

node/node.go
kehiy: I did that but there is an issue, all stuff is fine but I got this warning:
```
the cancel function is not used on all paths (possible context leak)
```
I already called cancel in the stop node method.

kehiy:
@b00f I added some changes!
I think that what you want, you can review it now

node/node.go
Ja7ad: > I think we need to add cancel to the node structure and call it when the node is shutting down

1. Cancel function just use in parent function, don't pass to other layers.
2. I think don't keep context in structure.

network/network.go
```diff
@@ -26,7 +26,6 @@ type network struct {
 	// We should remove it from here and pass it as first argument of functions
 	// Adding these linter later:  contextcheck and containedctx
 	ctx            context.Context
-	cancel         func()
```
Ja7ad, line 29: don't remove cancel function.

```go
cancel context.CancelFunc
```

network/network_test.go
```diff
@@ -104,7 +106,8 @@ func readData(t *testing.T, r io.ReadCloser, len int) []byte {
 }
 
 func TestStoppingNetwork(t *testing.T) {
-	net, err := NewNetwork(testConfig())
+	ctx := context.Background()
```
Ja7ad, line 109: for test you can use `context.ToDo`

node/node.go
Ja7ad: > 

Stop method used in any there for cancellation?

Ja7ad: > I did that but there is an issue, all stuff is fine but I got this warning:
> 
> ```
> the cancel function is not used on all paths (possible context leak)
> ```
> 
> I already called cancel in the stop node method.

`Stop` method use in any there?

Ja7ad: > I did that but there is an issue, all stuff is fine but I got this warning:
> 
> ```
> the cancel function is not used on all paths (possible context leak)
> ```
> 
> I already called cancel in the stop node method.

Has the Stop method been used anywhere?

kehiy: I don't think so.
where should I use it?

network/network_test.go
kehiy: is that the just thing I should change?

Ja7ad: > is that the just thing I should change?

Yes

node/node.go
Ja7ad: in GUI: CTA Exit
in CLI: on interrupt signal

kehiy: ok
I will add it

network/network_test.go
kehiy: ok, I will change it too
should I remove defer func for cancel() in node.go?

Ja7ad: > ok, I will change it too should I remove defer func for cancel() in node.go?

Yes

kehiy:
@b00f should we close this and follow the steps I provided and make a new pr?
or we can resolve them here?

kehiy:
> Better to have a new PR. Thanks

okay, after the script, I will send a new PR for it.

OmidHekayati:
Personally, I don't like using `Context` to solve when the requirements are just about object life cycles!

As [discuss here](https://github.com/pactus-project/pactus/discussions/544) suggest having three methods in all struct as an object to solve the object life cycle. When we have the `Deinit` method, anyone can easily know how an object de-initializes and reduce magic codes!

For a more deep discussion, I think it is better to have a meeting if you agree with the overall idea.

b00f:
> For a more deep discussion, I think it is better to have a meeting if you agree with the overall idea.

We have some DEV-meeting from time to time with the development team. There's no specific time for these meetings, but we can discuss it at the next DEV-meeting. 

Thanks for bringing it up @OmidHekayati

PR #584, opened 2023-07-09
fix: adding block hash to peer

b00f:
## Description

This PR includes two improvements. First, it adds the block hash to the peer structure. This helps us to find out both the latest block height and block hash of a peer. Second, get random votes based on the round. This change will help nodes retrieve the votes more quickly.


amirvalhalla:
@b00f  thanks, auto-merged enabled

b00f:
@amirvalhalla you need to approve it first

amirvalhalla (approved):

amirvalhalla:
@b00f 
done

Issue #585, opened 2023-07-11
Consensus halted


b00f:
## Summary

We encountered a serious issue on Testnet that resulted in a halt to the consensus. An issue within the consensus algorithm led to a fork in the blockchain. This fork halted the consensus mechanism, preventing it from advancing.
The Pactus consensus algorithm prioritizes [safety over (weaker) liveness](https://b00f.github.io/blockchain/consensus_problem), meaning that it is designed to prevent forks, and if one occurs, the system will halt.
However, a flaw in the consensus algorithm resulted in the creation of two blocks with valid certificates at the same height, leading to an unexpected fork in Pactus.

### What happened?

To explain the situation, let's say we have 4 peers: $P_A$, $P_B$, $P_C$, $P_D$.
$P_A$, $P_B$ and $P_C$ have very poor and unstable internet connections (for instance, they are geographically distant from each other and receive messages with significant delay). However, peer $P_D$ has a good connection.

In one height, $P_A$ proposes block $B_1$. The other peers then vote for this block, move to the `precommit` phase, and cast their `precommit` votes. However, while $P_A$, $P_B$, and $P_C$ are awaiting `precommit` votes from the other peers, their timers expire, leading them to transition to the `change-proposer` phase. In the following round, $P_B$ proposes block $B_2$. This time, block $B_2$ secures sufficient quorum and is announced along with a valid certificate.

The issue arises when $P_D$ (the peer with a good connection) receives all `precommit` votes for $B_1$ from $P_A$, $P_B$, and $P_C$. As the vote reaches a quorum, it's able to create a valid certificate for $B_1$ as well.  Consequently, we end up with two valid blocks, $B_1$ and $B_2$, at the same height, which should not occur. 

| Peer | prepare votes <br> (round 0) | precommit votes <br> (round 0) | prepare votes <br> (round 1) | precommit votes <br>(round 1) |
|---|---|---|---|---|
|$P_A$| $B_1: \set{A, B, C}$| $B_1: \set{A}$ | $B_2: \set{A, B, C}$| $B_2: \set{A, B, C}$ |
|$P_B$| $B_1: \set{A, B, C}$| $B_1: \set{A}$ | $B_2: \set{A, B, C}$| $B_2: \set{A, B, C}$ |
|$P_C$| $B_1: \set{A, B, C}$| $B_1: \set{A}$ | $B_2: \set{A, B, C}$| $B_2: \set{A, B, C}$ |
[...]

PR #589, opened 2023-07-14
docs: updating TLA+ spec

b00f:
## Description

To enter the "change-proposer" phase, a non-faulty node requires a quorum of "CHANGE-PROPOSER" votes. This ensures that non-faulty nodes do not transition into the "change-proposer" state too soon. 
Additionally, the Liveness property checked to make sure "eventually" a block will be committed.

## Related issue(s)

- Spec for fixing #585


nagarajmanjunath:
Concerning "quorum", is there a requisite minimum number of nodes that need to be verified? If this is the case, what is the estimated probability?

b00f:
> is there a requisite minimum number of nodes that need to be verified?

Here, 'quorum' refers to a supermajority, which is more than two-thirds (2/3+) of the votes. 
In committee, each vote is weighted according to the stake. Therefore, more than two-thirds of the committee's stake should vote for the proposed block.

consensus/spec/Pactus.pdf
Ja7ad, line 1: @b00f I check this algorithm and list some tips and pitfalls for double check:

**Tips:**

1. **Fault Tolerance:** The algorithm nicely incorporates Byzantine fault tolerance. It is designed to work even if up to `f` nodes are faulty in a system of `3f + 1` nodes. This is a typical characteristic of PBFT-based algorithms and is a good practice.

2. **Use of Helper Functions:** Helper functions such as `IsProposer`, `HasPrepareQuorum`, `HasPrecommitQuorum`, `HasChangeProposerQuorum`, etc., are used to encapsulate logic that checks certain conditions. This is good for code readability and maintainability.

3. **Message Broadcasting:** The functions like `SendMsg`, `SendProposal`, `SendPrepareVote`, etc., are designed to broadcast messages to all nodes. This ensures that all non-faulty nodes are able to receive every valid message.

**Potential Pitfalls:**

1. **Validity of Messages:** The `SendMsg` function does not check the validity of the message before broadcasting it. It only checks whether the current height is committed or not. This could potentially allow invalid messages to be broadcasted.

2. **Faulty Node Checks:** The checks for faulty nodes (`IsFaulty` function) are based on the index of the nodes. There seems to be an assumption that certain node indices are faulty, which may not always be the case in a real-world scenario.

3. **State Transitions:** There are numerous state transition functions like `NewHeight`, `Propose`, `Prepare`, etc. Each of these functions changes the state of nodes based on certain conditions. If these conditions are not met correctly, the state transitions could fail. 

4. **Consensus Completion:** It is not clear how the algorithm determines that consensus has been reached. Typically, in PBFT, consensus is reached when a client's request has been executed and a reply has been sent to the client by `f+1` non-faulty nodes.

5. **Error Handling:** There doesn't seem to be any error handling mechanisms in place. This could lead to unexpected behavior if something goes wrong.

b00f: Thanks @Ja7ad .

PR #595, opened 2023-07-17
refactor: replace logrus with zerolog

### LINKED ISSUE #587: Replacing Logrus with zerolog
Currently, [Logrus](https://github.com/sirupsen/logrus) is being used for logging. When certain levels like Trace are off, program still evaluates these log messages, leading to potential performance issues. For example:

```go
logger.Trace("Processing took: ", complexFunction())
```

In this case, `complexFunction()` is evaluated even if Trace level logging is off, adding unnecessary overhead.
An alternative approach is to use the [zerolog](https://github.com/rs/zerolog) library which provides more control over logging overhead:

```go
log.Trace().Msg("this message may not be logged depending on level", complexFunction())
```

In this case, `complexFunction()` would only be evaluated when the trace log level is enabled, which could improve performance by reducing unnecessary computations and log object creation.

For a performance comparison between zerolog and logrus, check out this [benchmark](http://hackemist.com/logbench/).

kehiy:
## Description
I replaced Logrus with the Zerolog package.

## Related issue(s)

- Fixes #587

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,8 +29,6 @@ require (
 	google.golang.org/protobuf v1.30.0
 )
 
-require github.com/mattn/go-colorable v0.1.12 // indirect
-
 require (
 	github.com/Microsoft/go-winio v0.5.2 // indirect
 	github.com/benbjohnson/clock v1.3.0 // indirect
@@ -88,6 +86,7 @@ require (
 	github.com/libp2p/go-yamux/v4 v4.0.0 // indirect
 	github.com/libp2p/zeroconf/v2 v2.2.0 // indirect
 	github.com/marten-seemann/tcp v0.0.0-20210406111302-dfbc87cc63fd // indirect
+	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.18 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.4 // indirect
 	github.com/miekg/dns v1.1.53 // indirect
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 	github.com/pelletier/go-toml v1.9.5
 	github.com/pkg/errors v0.9.1
 	github.com/rakyll/statik v0.1.7
-	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/tyler-smith/go-bip39 v1.1.0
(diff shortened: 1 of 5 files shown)


kehiy:
@b00f can you review this?

go.mod
```diff
@@ -30,6 +29,8 @@ require (
 	google.golang.org/protobuf v1.30.0
 )
 
+require github.com/mattn/go-colorable v0.1.12 // indirect
```
b00f, line 32: Why two requires? I think you can add it in the below lines

util/logger/logger.go
```diff
@@ -130,97 +123,112 @@ func keyvalsToFields(keyvals ...interface{}) logrus.Fields {
...
 }
 
-func (l *Logger) log(level logrus.Level, msg string, keyvals ...interface{}) {
-	if l.logger.IsLevelEnabled(level) {
-		l.withDefaultFields().WithFields(keyvalsToFields(keyvals...)).Log(level, msg)
-	}
-}
-
-func (l *Logger) With(keyvals ...interface{}) *logrus.Entry {
+func (l *Logger) With(keyvals ...interface{}) *zerolog.Event {
```
b00f, line 157: Do we really need this function? If not, can we delete it?

util/logger/logger.go
```diff
@@ -130,97 +123,112 @@ func keyvalsToFields(keyvals ...interface{}) logrus.Fields {
...
 	fields := keyvalsToFields(keyvals...)
-	return l.logger.WithFields(fields)
+	return l.logger.Log().Fields(fields)
 }
 
 func (l *Logger) Trace(msg string, keyvals ...interface{}) {
-	l.log(logrus.TraceLevel, msg, keyvals...)
+	l.Log(zerolog.TraceLevel, msg, keyvals...)
```
b00f, line 163: I am wondering is it possible to have something like:
```
l.logger.Trace().Msgf(msg, eyvals...)
```

We get two points:
1- No need to convert keyvals to fields
2- Get rid of the switch-case

b00f:
@kehiy  Some tests are failed 😔.

util/logger/logger.go
kehiy: what should I do for this?

kehiy:
> @kehiy Some tests are failed pensive.

the logger tests are fine!
but it seems mu changes made some side effects, where? i'm trying to find it.

go.mod
kehiy: I used `go mod tidy` 
I will add it to the first one.

util/logger/logger.go
b00f: If it's only for testing, we can remove it. We haven't needed it yet, and probably won't in the future.

kehiy: ok, I will remove it

PR #599, opened 2023-07-25
fix: change Buf to Protoc for proto generating

### LINKED ISSUE #590: No fields for transaction Fee

Cannot find `transaction fee` details in API or gRPC Call

We need to create/add API/fee to calculate the total transaction fee.

ZigBalthazar:
Hi @b00f 

## Description

based on our discussion about `Buf `and `Protoc `in `./Makefile` in Discord, I fixed the Buf problem with replace it with `Protoc `.

the `make proto` command did not work correctly and I tried to fix it

## Related issue(s)

- Fixes #590 (.proto problems)


kehiy:
@ZigBalthazar I think we still need the removed files in `.py`, `.js`, and ...

PR #607, opened 2023-07-28
feat: add reason to BlockResponse messages

### LINKED ISSUE #526: Sync: Adding rejection reason to blocks response message

When a node receives a [blocks request](https://github.com/pactus-project/pactus/blob/01d155b09c7db3ab279668f7973f7ebc9194cc89/sync/bundle/message/blocks_request.go#L9) message, it processes the message and generates a [blocks response](https://github.com/pactus-project/pactus/blob/01d155b09c7db3ab279668f7973f7ebc9194cc89/sync/bundle/message/blocks_response.go#L13). Each response has a code, the [ResponseCode](https://github.com/pactus-project/pactus/blob/01d155b09c7db3ab279668f7973f7ebc9194cc89/sync/bundle/message/message.go#LL9C3-L9C3), which informs the sender about how their request was handled. There are situations that a request may be rejected, such as when a non-existing blocks is requested. In these situations, the response code is set to 'Rejected'. However, the problem is that we don't provide any explanation to the sender about why their request was rejected.

A useful change would be to include an additional field in the response message to explain why the request was rejected. We could modify the `BlocksResponseMessage` structure like this:

```go
type BlocksResponseMessage struct {
	ResponseCode    ResponseCode       `cbor:"1,keyasint"`
	SessionID       int                `cbor:"2,keyasint"`
	From            uint32             `cbor:"3,keyasint"`
	BlocksData      [][]byte           `cbor:"4,keyasint"`
	LastCertificate *block.Certificate `cbor:"6,keyasint"`
	Reason          string             `cbor:"7,keyasint"`
}
```
In addition, we could simplify things by merging `ResponseCodeBusy` with `ResponseCodeRejected`, and use the reason field to state "too many requests" when that's the case.

~~### backward compatibility~~

~~Make sure that the other response codes remain the same. This will ensure that our changes are compatible with older versions. It's also important to test these changes on the Testnet.~~

amirvalhalla:
## Description

add reason of rejection to block response

## Related issue(s)

- Fixes #526

### DIFF
--- a/sync/bundle/bundle_test.go
+++ b/sync/bundle/bundle_test.go
@@ -39,7 +39,8 @@ func TestMessageCompress(t *testing.T) {
 		d, _ := b.Bytes()
 		blocksData = append(blocksData, d)
 	}
-	msg := message.NewBlocksResponseMessage(message.ResponseCodeBusy, 1234, 888, blocksData, nil)
+	msg := message.NewBlocksResponseMessage(message.ResponseCodeRejected, message.ResponseCodeRejected.String(),
+		1234, 888, blocksData, nil)
 	bdl := NewBundle(ts.RandomPeerID(), msg)
 	bs0, err := bdl.Encode()
 	assert.NoError(t, err)
--- a/sync/bundle/bundle_test.go
+++ b/sync/bundle/bundle_test.go
@@ -39,7 +39,7 @@ func TestMessageCompress(t *testing.T) {
 		d, _ := b.Bytes()
 		blocksData = append(blocksData, d)
 	}
-	msg := message.NewBlocksResponseMessage(message.ResponseCodeRejected, message.ResponseCodeRejected.String(),
+	msg := message.NewBlocksResponseMessage(message.ResponseCodeOK, message.ResponseCodeOK.String(),
 		1234, 888, blocksData, nil)
 	bdl := NewBundle(ts.RandomPeerID(), msg)
 	bs0, err := bdl.Encode()
--- a/sync/handler_blocks_request.go
+++ b/sync/handler_blocks_request.go
@@ -24,7 +24,7 @@ func (handler *blocksRequestHandler) ParseMessage(m message.Message, initiator p
 
 	if handler.peerSet.NumberOfOpenSessions() > handler.config.MaxOpenSessions {
 		handler.logger.Warn("we are busy", "message", msg, "pid", initiator)
-		response := message.NewBlocksResponseMessage(message.ResponseCodeBusy,
+		response := message.NewBlocksResponseMessage(message.ResponseCodeRejected, message.ResponseCodeRejected.String(),
 			msg.SessionID, 0, nil, nil)
 		handler.sendTo(response, initiator, msg.SessionID)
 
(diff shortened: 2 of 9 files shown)


b00f:
The main idea behind this PR is to provide a message to peers once their request is rejected. There are several reasons why a request might be rejected. For example, the peer is not known ([here](https://github.com/pactus-project/pactus/blob/6c36120f5f1958f58412c3ccb63ba0b1671bbcf4/sync/handler_blocks_request.go#L36)), or they request a block that we don't have ([here](https://github.com/pactus-project/pactus/blob/6c36120f5f1958f58412c3ccb63ba0b1671bbcf4/sync/handler_blocks_request.go#L50)), etc.

In my opinion, we should not consider a rejected request as a failure(or byzantine). It might be, but we can't consider it as such. Consequently, our current practice of increasing the failure counter is incorrect. In this context, we could merge the `Busy` and `Rejected` codes. I'm sorry, I didn't notice earlier that we were adding to the failure counter.

Please note that a reason could be added as a new parameter for the `NewBlocksResponseMessage` function.

amirvalhalla:
> The main idea behind this PR is to provide a message to peers once their request is rejected. There are several reasons why a request might be rejected. For example, the peer is not known ([here](https://github.com/pactus-project/pactus/blob/6c36120f5f1958f58412c3ccb63ba0b1671bbcf4/sync/handler_blocks_request.go#L36)), or they request a block that we don't have ([here](https://github.com/pactus-project/pactus/blob/6c36120f5f1958f58412c3ccb63ba0b1671bbcf4/sync/handler_blocks_request.go#L50)), etc.
> 
> In my opinion, we should not consider a rejected request as a failure(or byzantine). It might be, but we can't consider it as such. Consequently, our current practice of increasing the failure counter is incorrect. In this context, we could merge the `Busy` and `Rejected` codes. I'm sorry, I didn't notice earlier that we were adding to the failure counter.
> 
> Please note that a reason could be added as a new parameter for the `NewBlocksResponseMessage` function.

request changes resolved, please review.

sync/bundle/bundle_test.go
```diff
@@ -39,7 +39,8 @@ func TestMessageCompress(t *testing.T) {
 		d, _ := b.Bytes()
 		blocksData = append(blocksData, d)
 	}
-	msg := message.NewBlocksResponseMessage(message.ResponseCodeBusy, 1234, 888, blocksData, nil)
+	msg := message.NewBlocksResponseMessage(message.ResponseCodeRejected, message.ResponseCodeRejected.String(),
```
b00f, line 42: For this particular test, `TestMessageCompress`, It's better to set the response code to OK.

sync/handler_blocks_request.go
```diff
@@ -24,7 +24,7 @@ func (handler *blocksRequestHandler) ParseMessage(m message.Message, initiator p
 
 	if handler.peerSet.NumberOfOpenSessions() > handler.config.MaxOpenSessions {
 		handler.logger.Warn("we are busy", "message", msg, "pid", initiator)
-		response := message.NewBlocksResponseMessage(message.ResponseCodeBusy,
+		response := message.NewBlocksResponseMessage(message.ResponseCodeRejected, message.ResponseCodeRejected.String(),
```
b00f, line 27: The reason should set to: "busy", or "too many request".
We want to inform the caller, why the request is rejected.

b00f (changes requested):

sync/handler_blocks_request.go
amirvalhalla: message of `message.ResponseCodeRejected.String()` is "too many requests". so I don't think so that we should change this section.

b00f (approved):

PR #608, opened 2023-07-29
fix!: updating sortition executor

themantre:
## Description

This PR introduces changes to the sortition executor to address several issues:

- It fixes a bug that was causing duplicate sortition transactions.
- It corrects the calculation used to determine whether a proposer has proposed a block or not.
- The `LastJoinedHeight` replaced by  `LastSortitionHeight`

BREAKING CHANGE


execution/executor/sortition.go
```diff
@@ -31,27 +31,32 @@ func (e *SortitionExecutor) Execute(trx *tx.Tx, sb sandbox.Sandbox) error {
...
-		// before entering into the committee
-		// In non-strict mode we don't check the sequence number
+		// It is possible for a validator to generate multiple sortition transactions
+		// before entering the committee.
+		// In non-strict mode, we do not check the sequence number.
 		if val.Sequence()+1 != trx.Sequence() {
 			return errors.Errorf(errors.ErrInvalidSequence,
 				"expected: %v, got: %v", val.Sequence()+1, trx.Sequence())
 		}
+		// Check for the duplicated sortition transactions
```
b00f, line 47: Better comment: "// Check for the duplicated or expired sortition transactions"

types/account/account.go
```diff
@@ -70,11 +70,13 @@ func (acc *Account) AddToBalance(amt int64) {
 	acc.data.Balance += amt
 }
```
b00f, line 71: Revert this file

www/http/blockchain.go
```diff
@@ -32,7 +32,7 @@ func (s *Server) BlockchainHandler(w http.ResponseWriter, _ *http.Request) {
 		tm.addRowInt("Sequence", int(val.Sequence))
 		tm.addRowAmount("Stake", val.Stake)
 		tm.addRowInt("LastBondingHeight", int(val.LastBondingHeight))
-		tm.addRowInt("LastJoinedHeight", int(val.LastJoinedHeight))
```
b00f, line 35: Make it a function.

b00f (changes requested):

types/validator/validator.go
```diff
@@ -131,8 +131,8 @@ func (val *Validator) IncSequence() {
 }
 
 // UpdateLastJoinedHeight updates the last height at which the validator joined the committee.
-func (val *Validator) UpdateLastJoinedHeight(height uint32) {
-	val.data.LastJoinedHeight = height
+func (val *Validator) UpdateLastSortitionHeight(height uint32) {
```
amirvalhalla, line 134: what is your opinion about below idea?
should we use mutex for `UpdateLastSortitionHeight` method? because during our update maybe `LastSortitionHeight` method call so the result of `LastSortitionHeight` is not correct.

amirvalhalla: @b00f 
@themantre 

what is your opinion?

themantre: Adding a locking mechanism such as a mutex to low-level types like `account` or `validator` can have some side effects:

-  It could slow down the system, as these types are widely used in other packages.
- It can result in many locks, making the system difficult to debug.

A better approach is to ensure that the main packages are thread-safe and that resources are not shared between them. 
When sharing is unavoidable, we pass a copy of the underlying object to prevent shared resource usage.
As an example look the [validator](https://github.com/pactus-project/pactus/blob/6c36120f5f1958f58412c3ccb63ba0b1671bbcf4/sandbox/sandbox.go#L149) function in sandbox and the way we clone the validator objects there.

amirvalhalla (approved):
@themantre  thanks

kehiy:
@themantre LGTM

kehiy (approved):

b00f (approved):

PR #609, opened 2023-07-29
refactor: Using Generics for calculating Min and Max for numeric type #604

### LINKED ISSUE #604: Using Generics for calculating Min and Max for numeric type
Inside the [util](https://github.com/pactus-project/pactus/blob/dcf224206e15f1c27ad6c8d985f0bf94aefd98be/util/utils.go#L27C1-L27C1) package, we have some functions for calculating the minimum and maximum values for `int64` and `int32`. We can refactor these functions and replace them with generic functions, like:

```go
func Max[T constraints.Integer](a, b T) T {
	if a < b {
		return b
	}
	return a
}
```

By using the new `Max` function, we can update the [Len](https://github.com/pactus-project/pactus/blob/dcf224206e15f1c27ad6c8d985f0bf94aefd98be/sync/cache/cache.go#L64) function in the sync/cache package to:

```go
return util.Max(c.blocks.Len(), c.cetrs.Len())
```

sadaghiani:
## Description

Duplicate codes are reduced and the performance is as follows.

```
cpu: Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
BenchmarkMaxGeneric-4           1000000000               0.3357 ns/op          0 B/op          0 allocs/op
BenchmarkMax-4                  1000000000               0.3450 ns/op          0 B/op          0 allocs/op
PASS
```

## Related issue(s)

- Fixes #604


themantre:
Interesting @sadaghiani 
I am impressed by the benchmark result.
I can see the generic types have slightly better performance.  0.3357 compare to  0.3450 (+0.0093).

themantre (approved):

PR #618, opened 2023-07-31
feat: adding Sent and Received bytes per message metrics for peers

### LINKED ISSUE #605: Add metrics for Peers per message type
Similar to issue #594, we can add metrics per peer. The `sentBytes` and `receivedBytes` can be added for the [Peer](https://github.com/pactus-project/pactus/blob/dcf224206e15f1c27ad6c8d985f0bf94aefd98be/sync/peerset/peer.go#L18C1-L18C1) structure.

The only complexity here is that we should not increase the number of `sentBytes` on gossiping messages. To address this, we can add a new parameter to the `IncreaseTotalSentBytesCounter` function to obtain a pointer to the destination `peer_id`. If `peer_id` is nil, it is a broadcasting message.

Is there a better solution?

ZigBalthazar:
## Description

Regarding issue #605, I have made some updates. I have added "SentBytes" and "ReceivedBytes" to "peer" and made changes to the "IncreaseReceivedBytesCounter" and "IncreaseSentBytesCounter" functions. Additionally, I have modified the "sendTo" and "broadcast" functions and updated the Protos.

 ## Related issue(s)

- Fixes #605


ZigBalthazar:
I made a mistake while managing my local git branches, and this resulted in 31 commits. 🥲

sync/sync.go
```diff
@@ -327,7 +327,9 @@ func (sync *synchronizer) sendTo(msg message.Message, to peer.ID, sessionID int)
 			sync.logger.Info("sending bundle to a peer", "bundle", bdl, "to", to)
 			sync.peerSet.IncreaseSendSuccessCounter(to)
 		}
-		sync.peerSet.IncreaseSentBytesCounter(msg.Type(), len(data))
+
+		// pid := sync.SelfID()
```
b00f, line 331: Please remove this commented line

b00f (changes requested):

ZigBalthazar:
Hi @b00f 

Could you kindly provide me with your feedback?

b00f (approved):

PR #619, opened 2023-08-01
feat: firewall config contain trusted peers and only validator

### LINKED ISSUE #205: firewall: Only accept messages from Validators
Add new functionality to firewall to only accept messages from validators. 
We can check the initiator address to check if the message is initiated from a validator.
User can on/off this setting in firewall config file.

### LINKED ISSUE #207: Firewall: Only accept messages from trusted peer ids
We can add new functionality to the firewall to only accept messages that initiated from a list of trusted peer ids. 
User can update this list inside firewall config. If the list is empty, node accepts all incoming messages.

kehiy:
## Description
I added to parameter to the firewall config that allows the node owner only accept messages from validators and trusted peers.

## Related issue(s)

- Fixes #205
- Fixes #207


b00f:
Stale PR. Close it now

PR #623, opened 2023-08-06
fix: adding committers to the certificate

b00f:
## Description

This PR reverts #444.

While it may seem optimal to remove the committer list from the certificate, it's not practical for the following reasons:

1- It's logical for a certificate to include the list of its signers (or committers) alongside the signature.
2- Retrieving the committers of a certificate based on the current state of the blockchain isn't feasible without storing additional information.


amirvalhalla (approved):

kehiy (approved):
LGTM

PR #630, opened 2023-08-11
chore: fixing the Windows issue and test failure

themantre:
## Description

We have encountered a test failure in the system tests, which is caused by a random number for the stake that falls below the minimum threshold. In this pull request (PR), the stake amount has been fixed at 1000000. As a result, we should not experience any random failures in these tests.

Also, we ensure that `currentState` is not `nil` in the consensus.
Also, we ensure that `go1.20` is installed for Windows action.


kehiy (approved):
LGTM, which fixes the stake test and Windows build issue.

kehiy:
@themantre should we use `go1.21.0` for all of the projects?
I think it's better to be updated with the last version of Go.

PR #650, opened 2023-08-20
chore(linter): enable dupword linter

### LINKED ISSUE #649: dupword

kehiy:
## Description
I enabled `dupword` linter

## Related issue(s)
- Fixes #649

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -102,6 +102,10 @@ issues:
       linters:
         - funlen
 
+    - path: _test.go
+      linters:
+        - dupword
+
     - linters:
         - govet
       text: "shadow: declaration of \"err\" shadows"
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -45,9 +45,27 @@ linters:
     - nonamedreturns # Consider to enable it
     - nlreturn # Consider to enable it
     - ireturn  # Consider to enable it (??)
-    - dupword  # Consider to enable it
 
 linters-settings:
+  dupword:
```
b00f, line 50: Please revert the config file

b00f (changes requested):

.golangci.yml
kehiy: Do you mean I should remove the `dupword` and its setting from `linters-settings` ?

b00f: keep `dupword` but remove its settings

PR #664, opened 2023-08-22
chore: enable forceTypeAssert linter

### LINKED ISSUE #663: forcetypeassert

kehiy:
- Fixes #663

### DIFF
--- a/committee/committee.go
+++ b/committee/committee.go
@@ -90,8 +90,12 @@ func (c *committee) Update(lastRound int16, joined []*validator.Validator) {
 	}
 
 	sort.SliceStable(oldestFirst, func(i, j int) bool {
-		return oldestFirst[i].Value.(*validator.Validator).LastSortitionHeight() <
-			oldestFirst[j].Value.(*validator.Validator).LastSortitionHeight()
+		valI, valIok := oldestFirst[i].Value.(*validator.Validator)
+		valJ, valJok := oldestFirst[j].Value.(*validator.Validator)
+		if valIok && valJok {
+			return valI.LastSortitionHeight() < valJ.LastSortitionHeight()
+		}
+		return false
 	})
 
 	for i := 0; i <= int(lastRound); i++ {
@@ -164,7 +168,11 @@ func (c *committee) proposer(round int16) *validator.Validator {
 		}
 	}
 
-	return pos.Value.(*validator.Validator)
+	val, ok := pos.Value.(*validator.Validator)
+	if ok {
+		return val
+	}
+	return &validator.Validator{}
 }
 
 func (c *committee) Committers() []int32 {
@@ -204,8 +212,11 @@ func (c *committee) String() string {
 // iterate uses for easy iteration over validators in list.
 func (c *committee) iterate(consumer func(*validator.Validator) (stop bool)) {
 	for e := c.validatorList.Front(); e != nil; e = e.Next() {
-		if consumer(e.Value.(*validator.Validator)) {
-			return
+		val, ok := e.Value.(*validator.Validator)
+		if ok {
+			if consumer(val) {
+				return
+			}
 		}
 	}
 }
--- a/committee/committee.go
+++ b/committee/committee.go
@@ -172,7 +172,7 @@ func (c *committee) proposer(round int16) *validator.Validator {
 	if ok {
 		return val
 	}
-	return &validator.Validator{}
+	return nil
 }
 
 func (c *committee) Committers() []int32 {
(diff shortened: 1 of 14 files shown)


committee/committee.go
```diff
@@ -90,8 +90,12 @@ func (c *committee) Update(lastRound int16, joined []*validator.Validator) {
 	}
 
 	sort.SliceStable(oldestFirst, func(i, j int) bool {
-		return oldestFirst[i].Value.(*validator.Validator).LastSortitionHeight() <
-			oldestFirst[j].Value.(*validator.Validator).LastSortitionHeight()
+		valI, valIok := oldestFirst[i].Value.(*validator.Validator)
```
b00f, line 93: Can't we use generic here to avoid this type conversion?

committee/committee.go
```diff
@@ -204,8 +212,11 @@ func (c *committee) String() string {
 // iterate uses for easy iteration over validators in list.
 func (c *committee) iterate(consumer func(*validator.Validator) (stop bool)) {
 	for e := c.validatorList.Front(); e != nil; e = e.Next() {
-		if consumer(e.Value.(*validator.Validator)) {
-			return
+		val, ok := e.Value.(*validator.Validator)
```
b00f, line 215: Similar here

crypto/bls/private_key.go
```diff
@@ -143,5 +143,9 @@ func (prv *PrivateKey) PublicKey() crypto.PublicKey {
 }
 
 func (prv *PrivateKey) EqualsTo(right crypto.PrivateKey) bool {
-	return prv.fr.Equal(&right.(*PrivateKey).fr)
+	prvKey, ok := right.(*PrivateKey)
```
b00f, line 146: @amirvalhalla is looking for a way to avoid these casts, if it is possible.

b00f (commented):
I think we can refactor some logic to avoid these casts

util/testsuite/testsuite.go
```diff
@@ -291,7 +297,7 @@ func (ts *TestSuite) GenerateTestCertificate(blockHash hash.Hash) *block.Certifi
 	_, priv3 := ts.RandomBLSKeyPair()
 	_, priv4 := ts.RandomBLSKeyPair()
 
-	sigs := []*bls.Signature{
+	sigs := []*bls.Signature{ //nolint
```
b00f, line 300: Why?

kehiy: it makes tests more heavy and complex.

committee/committee.go
kehiy: I don't have idea now, I will search and try to make it done

util/testsuite/testsuite.go
b00f: Hopefully we can use generics here. @amirvalhalla any update?

kehiy: okay, I will wait for updates from him.
also, I can start work on this if you wish.

b00f: Yes, You may try as well

amirvalhalla: @b00f 
@kehiy 

I will work on make generic these methods until end of this weekend , please be patient :)

amirvalhalla: @b00f 
@kehiy 

Hi guys
Unfortunately, right now we can't use generics for `PrivateKey`, `PublicKey`,`Signature` etc... the reason is if we want to use generics we should change the structure of a lot of functions and interfaces to support generics.

e.g. in the `bls` package we implemented `PrivateKey` interface methods so if we want to use generics for `EqualsTo`  we should convert `PrivateKey` struct to `PrivateKey[T any]` this change makes some conflicts for other methods  like `PrivateKeyFromString` method, another reason is `Crypto.PrivateKey` interface is not generic so if we also change it to generic like `Crypto.PrivateKey[T any]` another issue is we can't cast it to another type.(if want to do that we should use embedded struct).

if generics are necessary which I think we should consider it, we should create another issue and discuss about what methods or packages should be generic then refactor our codes.

committee/committee.go
kehiy: fixed in :
- #667

committee/committee.go
kehiy: fixed in:
- #667

kehiy:
@b00f @amirvalhalla I'm going to close this issue for Conflicting files and after trying to fix some type assertions I will open a new PR for this linter.

PR #670, opened 2023-09-01
chore: fix unbond typo

kehiy:
## Description

from the previous PR, we changed `Unbond` to `Unbound`, but the `Unbond` phrase is correct, and changing it to `Unbound` needs a heavy change, so we decided to revert this word.

### DIFF
--- a/types/tx/payload/payload.go
+++ b/types/tx/payload/payload.go
@@ -13,7 +13,7 @@ const (
 	TypeTransfer  = Type(1)
 	TypeBond      = Type(2)
 	TypeSortition = Type(3)
-	TypeUnbound   = Type(4)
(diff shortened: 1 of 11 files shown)


types/tx/payload/payload.go
```diff
@@ -13,7 +13,7 @@ const (
 	TypeTransfer  = Type(1)
 	TypeBond      = Type(2)
 	TypeSortition = Type(3)
-	TypeUnbound   = Type(4)
+	TypeUnbond   = Type(4)
```
ZigBalthazar, line 16:
```suggestion
	TypeUnbond    = Type(4)
```

b00f (approved):

Issue #620, opened 2023-08-01
Hole punching and IPv6


kehiy:
## Describe the problem you'd like to have solved

we should find out the `IPv6` problem and try to use them if the device that runs the node able to use `IPv6`
already we should implement the `libp2p hole-punching` in `Pactus`. (also find a good way to test it)

### Alternatives and current workaround

for now, we use relay nodes, and also the full nodes can't use `IPv6` (the code seems to have supported it but there is an issue with this)

### Additional context

I will update this issue with more research and information till `testnet-2`

kehiy:
I think we need to attempt a manual test for hole-punch feature added in #697, but about IPv6 I think we should be interested in it anymore.

b00f:
Did you try to run metrics and see how it works?

kehiy:
hole punch or IPv6?

For hole-punch I think we can check at graphna(?) panel, I need access to it. it can how us the success rate.

PR #672, opened 2023-09-01
feat: adding udp protocol for network

b00f:
## Description

This PR adds a tests to make sure UDP  and IPv6 protocols are supported in the network module. 

## Related issue(s)

Maybe related to #620

### DIFF
--- a/network/errors.go
+++ b/network/errors.go
@@ -1,0 +1,33 @@
+package network
+
+import "fmt"
+
+// NotSubscribedError is returned when the peer is not subscribed to a
+// specific topic.
+type NotSubscribedError struct {
+	TopicID TopicID
+}
+
+func (e NotSubscribedError) Error() string {
+	return fmt.Sprintf("Not subscribed to the '%s' topic", e.TopicID.String())
+}
+
+// InvalidTopicError is returned when the Pub-Sub topic is invalid.
+type InvalidTopicError struct {
+	TopicID TopicID
+}
+
+func (e InvalidTopicError) Error() string {
+	return fmt.Sprintf("invalid topic: %s",
+		e.TopicID.String())
+}
+
+// LibP2PError is returned when an underlying libp2p operation encounters an error.
+type LibP2PError struct {
+	Err error
+}
+
+func (e LibP2PError) Error() string {
+	return fmt.Sprintf("libp2p error: %s",
+		e.Err.Error())
+}
--- a/network/errors.go
+++ b/network/errors.go
@@ -9,7 +9,7 @@ type NotSubscribedError struct {
 }
 
 func (e NotSubscribedError) Error() string {
-	return fmt.Sprintf("Not subscribed to the '%s' topic", e.TopicID.String())
+	return fmt.Sprintf("not subscribed to the '%s' topic", e.TopicID.String())
 }
 
 // InvalidTopicError is returned when the Pub-Sub topic is invalid.
@@ -18,7 +18,7 @@ type InvalidTopicError struct {
 }
 
 func (e InvalidTopicError) Error() string {
-	return fmt.Sprintf("invalid topic: %s",
+	return fmt.Sprintf("invalid topic: '%s'",
 		e.TopicID.String())
 }
 
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -341,7 +341,7 @@ func TestConnections(t *testing.T) {
 		{"/ip4/0.0.0.0/tcp/%d", "/ip4/0.0.0.0/tcp/0"},
 		{"/ip4/0.0.0.0/udp/%d/quic", "/ip4/0.0.0.0/udp/0/quic"},
 		{"/ip6/::/tcp/%d", "/ip6/::/tcp/0"},
-		{"/ip6/::/udp/%d/quic", "/ip6/::/udp/0/quic"},
+		// {"/ip6/::/udp/%d/quic", "/ip6/::/udp/0/quic"},
 	}
 
 	for i, test := range tests {
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -338,12 +338,12 @@ func TestConnections(t *testing.T) {
 		bootstrapAddr string
 		peerAddr      string
 	}{
-		{"/ip4/127.0.0.1/tcp/%d", "/ip4/127.0.0.1/tcp/0"},
+		{"/ip4/0.0.0.0/tcp/%d", "/ip4/0.0.0.0/tcp/0"},
 		{"/ip6/::/tcp/%d", "/ip6/::/tcp/0"},
-		{"/ip4/127.0.0.1/tcp/%d", "/ip6/::/tcp/0"},
-		{"/ip4/127.0.0.1/udp/%d/quic", "/ip6/::/udp/0/quic"},
+		{"/ip4/0.0.0.0/tcp/%d", "/ip6/::/tcp/0"},
+		{"/ip4/0.0.0.0/udp/%d/quic", "/ip6/::/udp/0/quic"},
 		{"/ip6/::/udp/%d/quic", "/ip6/::/tcp/0"},
-		{"/ip4/127.0.0.1/tcp/%d", "/ip4/127.0.0.1/udp/0/quic"},
+		{"/ip4/0.0.0.0/tcp/%d", "/ip4/0.0.0.0/udp/0/quic"},
 	}
 
 	for i, test := range tests {
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"testing"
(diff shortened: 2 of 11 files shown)


network/network_test.go
```diff
@@ -45,7 +46,7 @@ func makeTestNetwork(t *testing.T, conf *Config, opts []lp2p.Option) *network {
 	t.Helper()
 
 	net, err := newNetwork(conf, opts)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
ZigBalthazar, line 49: @b00f 
Why use assert instead of require here? Do you know if we can use the required in the following lines?

ZigBalthazar (changes requested):

network/network_test.go
b00f: We can sue both. In some cases (like above) if you use assert (and the net is nil) you will get panic in the next line.
But require stops there.

kehiy (approved):
everything looks good.

network/errors.go
```diff
@@ -0,0 +1,33 @@
...
+import "fmt"
+
+// NotSubscribedError is returned when the peer is not subscribed to a
+// specific topic.
+type NotSubscribedError struct {
+	TopicID TopicID
+}
+
+func (e NotSubscribedError) Error() string {
+	return fmt.Sprintf("Not subscribed to the '%s' topic", e.TopicID.String())
```
amirvalhalla, line 12: it would be better to whole of error message be lowercase (based on IDE linter warnings)

network/errors.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// InvalidTopicError is returned when the Pub-Sub topic is invalid.
+type InvalidTopicError struct {
+	TopicID TopicID
+}
+
+func (e InvalidTopicError) Error() string {
+	return fmt.Sprintf("invalid topic: %s",
```
amirvalhalla, line 21: I think it's better to have same syntax for returning errors, at `Error() string` method for `NotSubscribedError` struct you've implemented `return fmt.Sprintf("Not subscribed to the '%s' topic", e.TopicID.String())` which `%s` is in `' '` but in here it isn't and it would be better to have same syntax for here as other implementations.

ZigBalthazar (approved):
LGTM

network/network_test.go
```diff
@@ -21,7 +22,7 @@ func makeTestRelay(t *testing.T) host.Host {
 	t.Helper()
 
 	h, err := lp2p.New(
-		lp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/0"),
+		lp2p.ListenAddrStrings("/ip4/127.0.0.1/tcp/0"),
```
amirvalhalla, line 25: I think if we use `127.0.0.1` for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use `0.0.0.0` which work's like `127.0.0.1` in shell runners.

network/network_test.go
```diff
@@ -155,8 +158,8 @@ func TestNetwork(t *testing.T) {
 	confB := testConfig()
 	bootstrapPort := ts.RandInt32(9999) + 10000
 	confB.Listens = []string{
-		fmt.Sprintf("/ip4/0.0.0.0/tcp/%v", bootstrapPort),
-		fmt.Sprintf("/ip6/::/tcp/%v", bootstrapPort),
+		fmt.Sprintf("/ip4/127.0.0.1/tcp/%v", bootstrapPort),
```
amirvalhalla, line 161: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

network/network_test.go
```diff
@@ -171,8 +174,8 @@ func TestNetwork(t *testing.T) {
 	confP.EnableNAT = true
 	confP.Bootstrap.Addresses = bootstrapAddresses
 	confP.Listens = []string{
-		"/ip4/0.0.0.0/tcp/0",
-		"/ip6/::/tcp/0",
+		"/ip4/127.0.0.1/tcp/0",
```
amirvalhalla, line 177: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

network/network_test.go
```diff
@@ -186,8 +189,8 @@ func TestNetwork(t *testing.T) {
 	confM.RelayAddrs = relayAddrs
 	confM.Bootstrap.Addresses = bootstrapAddresses
 	confM.Listens = []string{
-		"/ip4/0.0.0.0/tcp/0",
-		"/ip6/::/tcp/0",
+		"/ip4/127.0.0.1/tcp/0",
```
amirvalhalla, line 192: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

network/network_test.go
```diff
@@ -201,8 +204,8 @@ func TestNetwork(t *testing.T) {
 	confN.RelayAddrs = relayAddrs
 	confN.Bootstrap.Addresses = bootstrapAddresses
 	confN.Listens = []string{
-		"/ip4/0.0.0.0/tcp/0",
-		"/ip6/::/tcp/0",
+		"/ip4/127.0.0.1/tcp/0",
```
amirvalhalla, line 207: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

network/network_test.go
```diff
@@ -215,8 +218,8 @@ func TestNetwork(t *testing.T) {
 	confX.EnableRelay = false
 	confX.Bootstrap.Addresses = bootstrapAddresses
 	confX.Listens = []string{
-		"/ip4/0.0.0.0/tcp/0",
-		"/ip6/::/tcp/0",
+		"/ip4/127.0.0.1/tcp/0",
```
amirvalhalla, line 221: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

network/network_test.go
```diff
@@ -334,3 +328,77 @@ func TestInvalidRelayAddress(t *testing.T) {
...
+func TestConnections(t *testing.T) {
+	t.Parallel() // run the tests in parallel
+
+	ts := testsuite.NewTestSuite(t)
+
+	tests := []struct {
+		bootstrapAddr string
+		peerAddr      string
+	}{
+		{"/ip4/127.0.0.1/tcp/%d", "/ip4/127.0.0.1/tcp/0"},
```
amirvalhalla, line 341: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

tests/main_test.go
```diff
@@ -89,7 +89,7 @@ func TestMain(m *testing.M) {
 		tConfigs[i].Sync.Firewall.Enabled = false
 		tConfigs[i].Network.EnableMdns = true
 		tConfigs[i].Network.NetworkKey = util.TempFilePath()
-		tConfigs[i].Network.Listens = []string{"/ip4/127.0.0.1/tcp/0"}
+		tConfigs[i].Network.Listens = []string{"/ip4/127.0.0.1/tcp/0", "/ip4/127.0.0.1/udp/0/quic"}
```
amirvalhalla, line 92: I think if we use 127.0.0.1 for tests it perfectly works on shell pipeline runners but in docker pipeline runners it would be better to use 0.0.0.0 which work's like 127.0.0.1 in shell runners.

amirvalhalla (changes requested):
@b00f 

network/errors.go
b00f: Thanks. Please check and resolve it

network/errors.go
b00f: Good point, Please check.

network/network_test.go
b00f: As I tested, in some case we had connection failure (probably firewall?) with `0.0.0.0`, but `127.0.01` for ipv4 and `::1` for ipv6 works fine.

b00f: Consider, the ports in these  tests are not going to expose. Like running tests inside docker.

network/network_test.go
b00f: Same as above ☝️

network/network_test.go
b00f: Same as above ☝️

network/network_test.go
b00f: Same as above ☝️

network/network_test.go
b00f: Same as above ☝️

network/network_test.go
b00f: Same as above ☝️

network/network_test.go
b00f: Same as above ☝️

tests/main_test.go
b00f: Same as above ☝️

amirvalhalla:
@b00f 
Thanks.

amirvalhalla (approved):

PR #673, opened 2023-09-02
feat: write logs into file

### LINKED ISSUE #495: Write logs to file

In our GUI application, we currently don't have logs for debugging purposes as we do not have access to stdout and stderr. 

Logs should be able to write into the file:

- The logs should be truncated based on either their size or date.
- An option to delete old log files should be provided.
- These settings can be configured within the configuration file.

Currently, there is no solution available unless the user runs the CLI application.

Related to #494

amirvalhalla:
## Description

write logs into files with rotation feature which provides optional parameters to configure log file based on time duration or file size 

## Related issue(s)

- Fixes #495


b00f:
@amirvalhalla 

Configuration can be simplified. 
User can just set if he want to write log to file and how much is the maximum size of the file.
So,

1- The directory always set to `logs` 
2- File name should set always to `pactus.log`
3- set `MaxAge`  to always zero. user can delete logs manually
We can set any other settings to default.

config/example_config.toml
```diff
@@ -130,6 +130,30 @@
 
 # `logger` contains configuration options for the logger.
 [logger]
+  # 'console_logging_enabled' indicates whether logs should stderr or stdout.
```
b00f, line 133: Please revert this file.

util/logger/config.go
```diff
@@ -1,14 +1,54 @@
...
 
+import "path/filepath"
+
+const (
+	LogDirectory = "logs"
+	LogFilename  = "pactus.log"
+)
+
 type Config struct {
+	// ConsoleLoggingEnabled Enable console logging
```
b00f, line 11: No new configuration for `log_to_file` and `file_rotation`. This file can be reverted.

util/logger/logger.go
```diff
@@ -89,14 +92,37 @@ func addFields(event *zerolog.Event, keyvals ...interface{}) *zerolog.Event {
 }
 
 func NewSubLogger(name string, obj fmt.Stringer) *SubLogger {
+	var writers []io.Writer
+
+	if getLoggersInst().config.ConsoleLoggingEnabled {
+		writers = append(writers, zerolog.ConsoleWriter{Out: os.Stderr})
```
b00f, line 98: Define two writers:
1- Console Writer
2- File Writer


File writer name is "pactus.log", the MaxSize is 100MB, Directory for backup is "logs" and out put is JSON (I am assuming that the default is JSON)

b00f (changes requested):

cmd/cmd.go
```diff
@@ -294,12 +294,12 @@ func CreateNode(numValidators int, chain genesis.ChainType, workingDir string,
 			return nil, nil, err
 		}
 
-		err = config.SaveTestnetConfig(confPath, numValidators)
+		err = config.SaveTestnetConfig(workingDir, confPath, numValidators)
```
b00f, line 297: There is no need to pass the "workingDir". The current directory is set as "workingDir", so any directory can be specified relative to "workingDir".

b00f (changes requested):

amirvalhalla:
@b00f 
we can't remove `WorkingDir` because if we use `os.GetWD` which returns current directory, the logs directory will generate for each package because when you run `os.GetWD` it return the current directory that code is executing.
so what is your suggestion? should we revert `WorkingDir` etc...

util/logger/logger.go
```diff
@@ -89,14 +93,45 @@ func addFields(event *zerolog.Event, keyvals ...interface{}) *zerolog.Event {
...
+	var fl *lumberjack.Logger
+	var logDirectory string
+	var logFilename string
+	maxLogSize := 100
+
+	// console writer
+	writers = append(writers, zerolog.ConsoleWriter{Out: os.Stderr})
+
+	// file writer
+	currentDirectory, err := os.Getwd()
```
b00f, line 106: This will change the current behavior of the project. By starting the node, the working directory changes to the user-defined directory. Look [here](https://github.com/pactus-project/pactus/blob/85f3063fa46ed4c892d1c7b8e7ec318e8cbe108d/cmd/daemon/start.go#L34) for details. Therefore, all logs should be saved inside that directory.

util/logger/logger.go
```diff
@@ -89,14 +93,45 @@ func addFields(event *zerolog.Event, keyvals ...interface{}) *zerolog.Event {
...
+	if err != nil {
+		goto ConsoleLogger
+	}
+
+	logDirectory = filepath.Join(currentDirectory, LogDirectory)
+	logFilename = filepath.Join(logDirectory, LogFilename)
+
+	if err = os.MkdirAll(logDirectory, 0o744); err != nil {
```
b00f, line 114: Based on [lumberjack](https://github.com/natefinch/lumberjack/blob/v2.0/README.md#type-logger) backup logs will be saved in the same directory. We don't need to create a directory for that.

b00f (changes requested):

util/logger/logger.go
amirvalhalla: @b00f 
I think there's no way to avoid passing `working dir` to `lumberjack` we should store user defined working directory then when the system will start node should read config and lumberjack will use this directory for logging.

b00f: The system always read from `<working-dir>/pactus.log` .
lumberjack package should not know where to store the log.

util/logger/config.go
```diff
@@ -1,5 +1,10 @@
 package logger
 
+const (
+	LogDirectory = "logs"
```
b00f, line 4: I don't think we need it anymore.

b00f (approved):

util/logger/config.go
```diff
@@ -1,5 +1,10 @@
...
+	LogFilename  = "pactus.log"
```
b00f, line 5: Please move this from `config.go` to `logger.go`
It is not something to be configured.

b00f (approved):

amirvalhalla:
@b00f 
I'll push last changes, tonight

b00f:
@amirvalhalla 
We can merge it now. No worries. I will fix it

amirvalhalla:
> @amirvalhalla We can merge it now. No worries. I will fix it

@b00f 
please let met push changes, everything is ok :)

PR #671, opened 2023-09-01
feat: implementing pip-0004 `indexing public key`

kehiy:
implementation of PIP-0004 which is defined here:
- PIP https://github.com/pactus-project/PIPs/pull/8

PR #675, opened 2023-09-04
chore: add pip reference for `Flags` in transaction data

kehiy:
## Description

added the `pip` reference of `Flags` in transaction data.


themantre:
We have referenced PIP in the PR #671 
The comments in the code should primarily focus on providing information about the code itself, rather than referencing external resources. 
Thank you for your understanding.

PR #681, opened 2023-09-10
chore: better logging by renaming err to error

themantre:
## Description

Renaming `err` to `error` improves logging in colorful logs.


amirvalhalla (approved):
@themantre 
thanks.

PR #683, opened 2023-09-10
feat: check protocol support before sending connect/disconnect event

### LINKED ISSUE #682: Stop sending connect/disconnect event to Relay nodes
We send a connect/disconnect event to `Relay` nodes each time from `notifee` service and `stream` protocol.
we should just send this events to `Pactus` nodes.

kehiy:
## Description
With this PR, we have 2 checks before sending a `connect` or `disconnect` event.
we make sure the node we are sending this event to, supports the `stream` protocol and It's a `Pactus` node.

## Related issue(s)
- Fixes #682


b00f (approved):

b00f:
@kehiy ,

As you know, when we connect to a peer, our handshaking process begins (`hello` -> `hello-ack`). The problem we encountered is that we receive connection notifications for relay peers as well. These relay peers don't support the stream protocol (`/pactus-testnet/stream/v1`), leading us to face some errors.

Your code attempts to determine the supported protocol by the connected peer. However, as I checked the LibP2P project, this event seems quite basic and is triggered when a node connects (or disconnects). They didn't negotiated their protocols yet. it's strange for me to me why they don't trigger an event after that process. 

Given the current situation, we cannot detect the supported protocol here, unless we wait or attempt to send some test data (perhaps junk data?).

For now, we must ensure we don’t close any connections if there's an error in the sync package. Let’s allow Libp2p to handle the connection and disconnection processes. It's not really good, but our options seem limited. Perhaps we can think out of box and find another solution?

PR #691, opened 2023-09-12
fix: restoring at the first block

b00f:
## Description

The genesis can have more than one account. This PR checks if the number of accounts in the store exceeds the number of accounts in the genesis. If it does, it's considered that the blockchain is no longer at the genesis height.


kehiy (approved):
everything looks good.
@b00f Thanks

PR #693, opened 2023-09-12
fix: ensure log rotation using tests

### LINKED ISSUE #666: Logger formating issue
There are two issues with formatting in the logger:

1- It appears we don't format the global logs. For example:
```
9:03PM DBG last info _state="{#0 ⌘ 000000000000 🕣 13.03.50}" committers=[0,1,2,3] state_root=9CFF11C4B393
{"level":"info","validator address":"pc1pjj2lrfsh7uewzpcn87xf4v7hy0leftgvar3p6m","reward address":"pc1pjj2lrfsh7uewzpcn87xf4v7hy0leftgvar3p6m","time":"2023-08-24T21:03:47+08:00","message":"consensus instance created"}
```

The first line is well formatted, but the second line, which is a global log, is not.

2- We should detect if the terminal doesn't support line editing, we don't format the logs. Example:
```
[90m9:09PM[0m [33mDBG[0m last info [36m_state=[0m"{#0 ⌘ 000000000000 🕣 13.09.10}" [36mcommitters=[0m[0,1,2,3] [36mstate_root=[0mD9E83BB17BF5
```

We should determine if the terminal supports line editing. If not, we shouldn't format the logs. To detect line editing support, we can use  [terminalSupported](https://github.com/pactus-project/pactus/blob/4622c3f58b9ba1caa8702923256f8354101247c3/cmd/cmd.go#L32C6-L32C23) 

We might consider moving this function to the `util` package.

b00f:
## Description

This PR:
- Enables writing to `os.Stderr` and to files for both global and sub-loggers.
- Ensures the working space remains unaffected during project testing.
- Introduces a test for log rotation.

## Related issue(s)

This PR also fixes #666

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -129,7 +129,7 @@ func SaveTestnetConfig(path string, numValidators int) error {
 	conf.GRPC.Listen = "[::]:50052"
 	conf.GRPC.Gateway.Enable = true
 	conf.GRPC.Gateway.Listen = "[::]:8080"
-	conf.HTTP.Enable = true
+	conf.HTTP.Enable = false
 	conf.HTTP.Listen = "[::]:80"
 	conf.Nanomsg.Enable = true
 	conf.Nanomsg.Listen = "tcp://127.0.0.1:40799"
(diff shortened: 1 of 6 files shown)


amirvalhalla (approved):
@b00f 
thanks

PR #697, opened 2023-09-13
feat: Adding Hole Punching to network

kehiy:
## Description

I added the `hole punching` config for the network.
**(After checking the success rate on Prometheus or Grafna we can close the related issue)**

## Related issue(s)

Related #620

### DIFF
--- a/config/example_config.toml
+++ b/config/example_config.toml
@@ -37,11 +37,6 @@
   # Default is true.
  ## enable_relay = false
 
-  # `enable_hole_punching` indicates whether Hole Punch service  should be enabled or not.
-  # Hole Punch service is an approach to NAT traversal.
-  # Default is true.
- ## enable_hole_punching = false
-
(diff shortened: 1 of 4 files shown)


config/example_config.toml
```diff
@@ -37,6 +37,11 @@
   # Default is true.
  ## enable_relay = false
 
+  # `enable_hole_punching` indicates whether Hole Punch service  should be enabled or not.
+  # Hole Punch service is an approach to NAT traversal.
+  # Default is true.
+ ## enable_hole_punching = false
```
b00f, line 43: I believe that when a user enables Relay, it can include HolePunching as well. This prevents conflicts between configs.
Consider a scenario where a user disables Relay but enables HolePunching.

kehiy: So, you mean we should merge the Relay config and hole punching config?

b00f (approved):

PR #704, opened 2023-09-19
feat: new API to get Public key by address

### LINKED ISSUE #449: Adding new API to get Publick key by address
## Describe the problem

Recently, we have indexed public keys by their addresses: #437 
This helps reduce the size of blocks by removing duplicated public keys. Now, we can add a new API in gRPC to get the public key associated with a specific address.

# Ideal Solution

We should add a new RPC function named `GetPublicKey` in `blockchain.proto`. It will accept an address as input and return the corresponding Public Key as output. If the public key is not indexed yet, it can return a `NotFound` error.

kehiy:
## Related issue(s)

- Fixes #449


node/node.go
```diff
@@ -78,7 +78,7 @@ func NewNode(genDoc *genesis.Genesis, conf *config.Config,
 	}
 
 	http := http.NewServer(conf.HTTP)
-	grpc := grpc.NewServer(conf.GRPC, state, sync, consMgr)
+	grpc := grpc.NewServer(conf.GRPC, state, store, sync, consMgr)
```
b00f, line 81: The reason we defined [Facade](https://github.com/pactus-project/pactus/blob/main/state/facade.go) is to encapsulated the store.
So please avoid it. instead define a function in state.Facade to get the Public Key

www/grpc/blockchain.go
```diff
@@ -222,6 +224,22 @@ func (s *blockchainServer) GetValidatorAddresses(_ context.Context,
...
+	req *pactus.GetPublicKeyRequest,
+) (*pactus.GetPublicKeyResponse, error) {
+	addr, err := crypto.AddressFromString(req.Address)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid account address: %v", err.Error())
+	}
+
+	publicKey, err := s.store.PublicKey(addr)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "account not found")
```
b00f, line 237:
```suggestion
		return nil, status.Errorf(codes.NotFound, "public key not found")
```

www/grpc/blockchain_test.go
```diff
@@ -264,6 +264,39 @@ func TestGetValidatorAddresses(t *testing.T) {
...
+
+	t.Run("Should return error for non-parsable address ", func(t *testing.T) {
+		res, err := client.GetPublicKey(tCtx,
+			&pactus.GetPublicKeyRequest{Address: ""})
+
+		assert.Error(t, err)
+		assert.Nil(t, res)
+	})
+
+	t.Run("Should return nil for non existing account ", func(t *testing.T) {
```
b00f, line 281:
```suggestion
	t.Run("Should return nil for non existing public key ", func(t *testing.T) {
```

www/grpc/blockchain_test.go
```diff
@@ -264,6 +264,39 @@ func TestGetValidatorAddresses(t *testing.T) {
...
+		res, err := client.GetPublicKey(tCtx,
+			&pactus.GetPublicKeyRequest{Address: ts.RandAddress().String()})
+
+		assert.Error(t, err)
+		assert.Nil(t, res)
+	})
+
+	t.Run("Should return account public key", func(t *testing.T) {
```
b00f, line 289:
```suggestion
	t.Run("Should return the public key", func(t *testing.T) {
```

www/grpc/grpc-gateway.config.yaml
```diff
@@ -25,6 +25,9 @@ http:
     - selector: pactus.Blockchain.GetBlockchainInfo
       get: "/v1/blockchain/info"
 
+    - selector: pactus.Blockchain.GetPublicKey
+      get: "/v1/blockchain/public_key/address/{address}"
```
b00f, line 29:
```suggestion
      get: "/v1/blockchain/public_keys/address/{address}"
```

Please consult with @Ja7ad

www/grpc/server.go
```diff
@@ -19,18 +20,20 @@ type Server struct {
...
 	grpc     *grpc.Server
 	state    state.Facade
+	store    store.Reader
 	sync     sync.Synchronizer
 	consMgr  consensus.ManagerReader
 	logger   *logger.SubLogger
 }
 
-func NewServer(conf *Config, state state.Facade, sync sync.Synchronizer,
+func NewServer(conf *Config, state state.Facade, store store.Reader, sync sync.Synchronizer,
```
b00f, line 29: Revert this change, as we discussed above.

www/http/http_test.go
```diff
@@ -37,6 +38,7 @@ func setup(t *testing.T) *testData {
 	ts := testsuite.NewTestSuite(t)
 
 	mockState := state.MockingState(ts)
+	mockStore := store.MockingStore(ts)
```
b00f, line 41: revert this file.

b00f: Please add the method to http server. It is helpful for debugging sometimes.

b00f (changes requested):

state/facade.go
```diff
@@ -50,4 +51,5 @@ type Facade interface {
 	Params() param.Params
 	Close() error
 	CalculateFee(amount int64, payloadType payload.Type) (int64, error)
+	GetPublicKey(addr crypto.Address) (*bls.PublicKey, error)
```
b00f, line 54:
```suggestion
	PublicKey(addr crypto.Address) (crypto.PublicKey, error)
```

state/mock.go
```diff
@@ -265,3 +266,7 @@ func (m *MockState) CalculateFee(_ int64, payloadType payload.Type) (int64, erro
 		return 0, errors.Errorf(errors.ErrInvalidTx, "unexpected tx type: %v", payloadType)
 	}
 }
+
+func (m *MockState) GetPublicKey(_ crypto.Address) (*bls.PublicKey, error) {
+	return &bls.PublicKey{}, nil
```
b00f, line 271:
```suggestion
	return m.store.PublicKey(addr)
```

state/state.go
```diff
@@ -734,3 +735,11 @@ func (st *state) CalculateFee(amount int64, payloadType payload.Type) (int64, er
 		return 0, errors.Errorf(errors.ErrInvalidTx, "unexpected tx type: %v", payloadType)
 	}
 }
+
+func (st *state) GetPublicKey(addr crypto.Address) (*bls.PublicKey, error) {
+	pubKey, err := st.store.PublicKey(addr)
+	if err != nil {
+		return &bls.PublicKey{}, err
```
b00f, line 742:
```suggestion
		return nil, err
```

b00f (changes requested):

PR #712, opened 2023-09-28
feat: implementing pip-11

b00f:
## Description

This PR implements the [PIP-11](https://pips.pactus.org/PIPs/pip-11).

### DIFF
--- a/crypto/bls/hdkeychain/extendedkey.go
+++ b/crypto/bls/hdkeychain/extendedkey.go
@@ -1,8 +1,8 @@
 package hdkeychain
 
 // References:
-//   [BIP32]: BIP0032 - Hierarchical Deterministic Wallets
-//   https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
+//  [PIP-11]: Deterministic key hierarchy for BLS-12381 curve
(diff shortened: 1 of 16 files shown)


themantre (approved):

crypto/bls/hdkeychain/extendedkey.go
```diff
@@ -140,11 +140,11 @@ func (k *ExtendedKey) Derive(index uint32) (*ExtendedKey, error) {
 	// child is hardened.
 	//
 	// For hardened children:
-	//   data (36 bytes) = parent_private_key (32 bytes)  || index (4 bytes)
+	//   0x00 || ser256(parentKey) || ser32(i)
```
amirvalhalla, line 143: at comments you just written about `point G2`  which is `0x00` but didn't write about `point G1` (0x01).
do am I mistake? or needed to add some comments about `point G1`

amirvalhalla (approved):

crypto/bls/hdkeychain/extendedkey.go
b00f: @amirvalhalla Good find.

I have updated the comments to enhance clarity.

PR #722, opened 2023-10-02
fix: fixing logger level issue

themantre:
## Description

This PR addresses the problem of the logger level config not being applied.


kehiy (approved):
@themantre  Thanks.

PR #723, opened 2023-10-02
fix: send query votes message, if there is no proposal yet

themantre:
## Description

Send query votes message if the query proposal timeout expires.


kehiy (approved):
@themantre Thanks.

PR #727, opened 2023-10-03
fix: resolve consensus halt caused by time discrepancy in network.

themantre:
## Description

If there is a time discrepancy in the network, even for a few seconds, there is a high chance that some validators missed votes or proposals. Prior to this PR, votes and proposals that did not belong to the current height were discarded.

This PR enhances the consensus with two improvements:
1. It immediately moves to the next height once a block is committed. This allows us to populate the votes from the next height before the `newHeight` timer expires.
2. In the `consensusManager`, we populate all votes and proposals belonging to upcoming heights and process them when the consensus moves to the new height.

Another improvement is the [PIP](https://pips.pactus.org/PIPs/pip-7) that prevents nodes with a high time discrepancy from joining the network.


consensus/manager.go
```diff
@@ -89,19 +99,81 @@ func (mgr *manager) MoveToNewHeight() {
...
 // SetProposal sets the proposal for all consensus instances.
-func (mgr *manager) SetProposal(proposal *proposal.Proposal) {
-	for _, cons := range mgr.instances {
-		cons.SetProposal(proposal)
+func (mgr *manager) SetProposal(p *proposal.Proposal) {
+	inst := mgr.getBestInstance()
+	curHeight, _ := inst.HeightRound()
+	switch {
+	case p.Height() < curHeight:
+		// discard the old proposal
```
amirvalhalla, line 168: @themantre 
for discarding the old proposal we should not do any operation?

themantre: No, just ignore them.

amirvalhalla (changes requested):
@themantre 

amirvalhalla (approved):

b00f (approved):

PR #730, opened 2023-10-05
fix: cache certificate by height

themantre:
## Description

This PR implements a more accurate approach to caching the certificate by its height.
It also removes the `height` parameter from the `CommitBlock` function, as the certificate already contains the `height` of the committing block.

### DIFF
--- a/state/facade.go
+++ b/state/facade.go
@@ -26,8 +26,8 @@ type Facade interface {
 	LastCertificate() *certificate.Certificate
 	UpdateLastCertificate(v *vote.Vote) error
 	ProposeBlock(valKey *bls.ValidatorKey, rewardAddr crypto.Address, round int16) (*block.Block, error)
-	ValidateBlock(block *block.Block) error
-	CommitBlock(block *block.Block, cert *certificate.Certificate) error
+	ValidateBlock(blk *block.Block) error
+	CommitBlock(blk *block.Block, cert *certificate.Certificate) error
 	CommitteeValidators() []*validator.Validator
 	IsInCommittee(addr crypto.Address) bool
 	Proposer(round int16) *validator.Validator
--- a/state/facade.go
+++ b/state/facade.go
@@ -27,7 +27,7 @@ type Facade interface {
 	UpdateLastCertificate(v *vote.Vote) error
 	ProposeBlock(valKey *bls.ValidatorKey, rewardAddr crypto.Address, round int16) (*block.Block, error)
 	ValidateBlock(block *block.Block) error
-	CommitBlock(height uint32, block *block.Block, cert *certificate.Certificate) error
+	CommitBlock(block *block.Block, cert *certificate.Certificate) error
 	CommitteeValidators() []*validator.Validator
 	IsInCommittee(addr crypto.Address) bool
 	Proposer(round int16) *validator.Validator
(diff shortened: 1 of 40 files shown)


state/facade.go
```diff
@@ -27,7 +27,7 @@ type Facade interface {
 	UpdateLastCertificate(v *vote.Vote) error
 	ProposeBlock(valKey *bls.ValidatorKey, rewardAddr crypto.Address, round int16) (*block.Block, error)
 	ValidateBlock(block *block.Block) error
-	CommitBlock(height uint32, block *block.Block, cert *certificate.Certificate) error
+	CommitBlock(block *block.Block, cert *certificate.Certificate) error
```
amirvalhalla, line 30: @themantre 
it would be better to use `blk` instead of `block` based on linter warning.

state/state.go
```diff
@@ -378,10 +378,11 @@ func (st *state) ValidateBlock(block *block.Block) error {
 	return st.executeBlock(block, sb)
 }
 
-func (st *state) CommitBlock(height uint32, block *block.Block, cert *certificate.Certificate) error {
+func (st *state) CommitBlock(block *block.Block, cert *certificate.Certificate) error {
```
amirvalhalla, line 381: @themantre 
same `blk` for here

store/interface.go
```diff
@@ -94,15 +94,15 @@ type Reader interface {
...
+	LastCertificate() *certificate.Certificate
 }
 
 type Store interface {
 	Reader
 
 	UpdateAccount(addr crypto.Address, acc *account.Account)
 	UpdateValidator(val *validator.Validator)
-	SaveBlock(height uint32, block *block.Block, cert *certificate.Certificate)
+	SaveBlock(block *block.Block, cert *certificate.Certificate)
```
amirvalhalla, line 105: @themantre 
same `blk` for here

store/store.go
```diff
@@ -89,18 +89,19 @@ func (s *store) Close() error {
 	return s.db.Close()
 }
 
-func (s *store) SaveBlock(height uint32, block *block.Block, cert *certificate.Certificate) {
+func (s *store) SaveBlock(block *block.Block, cert *certificate.Certificate) {
```
amirvalhalla, line 92: @themantre 
same `blk` for here

store/store.go
```diff
@@ -89,18 +89,19 @@ func (s *store) Close() error {
...
+	height := cert.Height()
 	reg := s.blockStore.saveBlock(s.batch, height, block)
 	for i, trx := range block.Transactions() {
 		s.txStore.saveTx(s.batch, trx.ID(), &reg[i])
 	}
 
 	// Save last certificate
-	w := bytes.NewBuffer(make([]byte, 0, 8+cert.SerializeSize()))
-	err := encoding.WriteElements(w, lastStoreVersion, height)
+	w := bytes.NewBuffer(make([]byte, 0, 4+cert.SerializeSize()))
```
amirvalhalla, line 103: @themantre 
it would be better if write some comments about what means `4` for capacity of buffer

amirvalhalla (changes requested):
@themantre 
thanks, please resolve request changes.

amirvalhalla (approved):
@themantre 
thanks, everything looks good.

PR #731, opened 2023-10-05
feat: implementing pip-7

amirvalhalla:
## Description

This PR implemented [pip-7](https://pips.pactus.org/PIPs/pip-7)

### DIFF
--- a/sync/bundle/message/hello_test.go
+++ b/sync/bundle/message/hello_test.go
@@ -56,16 +56,6 @@ func TestHelloMessage(t *testing.T) {
 		assert.GreaterOrEqual(t, m.MyTimeUnixMilli, myTimeUnixMilli)
 	})
 
-	t.Run("MyTimeUnixMilli of time1 is greater than hello message time", func(t *testing.T) {
-		m := NewHelloMessage(ts.RandPeerID(), "Alice", 100, 0, ts.RandHash(), ts.RandHash())
-		<-time.After(time.Millisecond)
-
-		time1 := time.Now()
-		myTimeUnixMilli := time1.UnixMilli()
-
-		assert.GreaterOrEqual(t, myTimeUnixMilli, m.MyTimeUnixMilli)
-	})
-
 	t.Run("Ok", func(t *testing.T) {
 		valKey := ts.RandValKey()
 		m := NewHelloMessage(ts.RandPeerID(), "Alice", 100, 0, ts.RandHash(), ts.RandHash())
--- a/sync/bundle/message/hello_test.go
+++ b/sync/bundle/message/hello_test.go
@@ -2,6 +2,7 @@ package message
 
 import (
 	"testing"
+	"time"
 
 	"github.com/pactus-project/pactus/crypto"
 	"github.com/pactus-project/pactus/crypto/bls"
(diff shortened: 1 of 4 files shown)


kehiy (approved):
@amirvalhalla Thanks, everything same as pip-7.

sync/bundle/message/hello_test.go
```diff
@@ -15,6 +16,17 @@ func TestHelloType(t *testing.T) {
 	assert.Equal(t, m.Type(), TypeHello)
 }
 
+func TestHelloMessageMyTimeUnix(t *testing.T) {
+	myTime := time.Now()
```
b00f, line 20: We can write a better test;
create time1 before NewHelloMessage(...)

Then time in HelloMessage should be GreaterOrEqual than `time1` and SmalerOrEqual Now.

b00f (changes requested):

sync/handler_hello.go
```diff
@@ -38,6 +39,14 @@ func (handler *helloHandler) ParseMessage(m message.Message, initiator peer.ID)
 		return handler.acknowledge(response, initiator)
 	}
 
+	if time.Since(msg.MyTime()) > 10*time.Second {
+		response := message.NewHelloAckMessage(message.ResponseCodeRejected,
+			fmt.Sprintf("hello message difference time is more than 10 seconds,"+
```
b00f, line 44: Messages might be slow to reach the node, so let's just say the times on the two nodes are different by more than 10 seconds: `time discrepancy exceeds 10 seconds`.

sync/handler_hello_test.go
```diff
@@ -44,8 +45,24 @@ func TestParsingHelloMessages(t *testing.T) {
 
 			assert.NoError(t, td.receivingNewMessage(td.sync, msg, pid))
 			td.checkPeerStatus(t, pid, peerset.StatusCodeBanned)
-			bundle := td.shouldPublishMessageWithThisType(t, td.network, message.TypeHelloAck)
-			assert.Equal(t, bundle.Message.(*message.HelloAckMessage).ResponseCode, message.ResponseCodeRejected)
+			bdl := td.shouldPublishMessageWithThisType(t, td.network, message.TypeHelloAck)
+			assert.Equal(t, bdl.Message.(*message.HelloAckMessage).ResponseCode, message.ResponseCodeRejected)
+		})
+
+	t.Run("Receiving Hello message from a peer. Difference of Hello Message Time with node time is a lot.",
```
b00f, line 52: We need two tests:
In the first test, `my_time` is: `Now() - 10 seconds`.
In the second test, `my_time` is: `Now() + 20 seconds`.

I purposely added an extra 10 seconds to the time to support running tests in slow CI/CD actions.

b00f (changes requested):

sync/bundle/message/hello_test.go
```diff
@@ -45,6 +46,26 @@ func TestHelloMessage(t *testing.T) {
...
+		time1 := time.Now()
+		myTimeUnixMilli := time1.UnixMilli()
+
+		m := NewHelloMessage(ts.RandPeerID(), "Alice", 100, 0, ts.RandHash(), ts.RandHash())
+
+		assert.LessOrEqual(t, m.MyTimeUnixMilli, time.Now().UnixMilli())
+		assert.GreaterOrEqual(t, m.MyTimeUnixMilli, myTimeUnixMilli)
+	})
+
+	t.Run("MyTimeUnixMilli of time1 is greater than hello message time", func(t *testing.T) {
```
b00f, line 59: This test is redundant. right? We almost cover every thing in above test.

b00f (approved):

b00f:
@amirvalhalla Looks good. Just one tests looks extra.

sync/bundle/message/hello_test.go
amirvalhalla: @b00f 
yes, I've forgot to remove this test, I'll fix it right now

b00f (approved):

b00f (approved):

PR #734, opened 2023-10-06
fix(consensus): anti-entroy mechanism for the consensus

b00f:
## Description

This PR added an Anti-Entropy mechanism to the consensus.
It also simplified the manager tests.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -107,8 +107,6 @@ func (sync *synchronizer) Start() error {
 	go sync.receiveLoop()
 	go sync.broadcastLoop()
 
-	sync.moveConsensusToNewHeight()
-
 	return nil
 }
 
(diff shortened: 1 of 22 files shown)


sync/sync.go
```diff
@@ -140,7 +138,10 @@ func (sync *synchronizer) sayHello(to peer.ID) error {
 	)
 	msg.Sign(sync.valKeys)
 
-	sync.peerSet.UpdateStatus(to, peerset.StatusCodeConnected)
+	peer := sync.peerSet.GetPeer(to)
+	if !peer.IsKnownOrTrusty() {
+		sync.peerSet.UpdateStatus(to, peerset.StatusCodeConnected)
```
amirvalhalla, line 143: @b00f 
why should we update status to `StatusCodeConnected` when peer is not Known or trusty?

amirvalhalla (changes requested):
@b00f 

sync/sync.go
b00f: If there is a connection to a "known" peer and the node restarts, the status is not changed to "connected".

amirvalhalla (approved):

PR #744, opened 2023-10-11
fix(network): async connection

b00f:
## Description

This PR introduces an asynchronous peer connection approach aimed at reducing node boot time.


amirvalhalla (approved):
@b00f 
thanks, everything sounds good.

PR #746, opened 2023-10-12
feat: add absentee votes to the certificate

amirvalhalla:
## Description

This PR adds absentees votes to the block certificate.

### DIFF
--- a/consensus/consensus_test.go
+++ b/consensus/consensus_test.go
@@ -71,6 +71,7 @@ func (o *OverrideStringer) String() string {
 
 func setup(t *testing.T) *testData {
 	t.Helper()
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -272,6 +272,9 @@ func (st *state) UpdateLastCertificate(v *vote.Vote) error {
 	defer st.lk.Unlock()
 
 	lastCert := st.lastInfo.Certificate()
+	// prevent race condition
+	cloneLastCert := lastCert.Clone()
```
b00f, line 276: For better performance, let's move this line to  301

themantre (approved):

PR #758, opened 2023-10-20
feat: added bootstrap node detection in config

### LINKED ISSUE #756: Detecting Bootstraper node
It is necessary to detect if a node is a bootstrap node. 
One way is by adding a boolean value to the config file. 
Another way is to read all bootstrap peers in the config, and if the node's peer ID is the same as one of the bootstrap nodes, it can be marked as a bootstrap node. In this case, we keep the config intact. 

After detecting the bootstrap node, we can mark it as a `NodeServer`. You can find an example of this implementation [here](https://github.com/filecoin-project/lotus/blob/14a7ae339fa57459283cd62ce8c0070b55a39c64/node/modules/lp2p/host.go#L87).

Note: We can also prevent the "Dial to self" error message.

melsonic:
## Description

Added a boolean value for detecting a bootstrap node in the config file.
## Related issue(s)
- Fixes #756

/claim #756

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,6 +41,7 @@ type Config struct {
 type NodeConfig struct {
 	NumValidators   int      `toml:"num_validators"` // TODO: we can remove this now
 	RewardAddresses []string `toml:"reward_addresses"`
+	BootstrapNode   bool 		 `toml:"bootstrap_node"`
 }
 
 func DefaultNodeConfig() *NodeConfig {


themantre:
@melsonic 
Thanks for your PR. We are going to detect the Bootstrap node without changing the config file.
We clearly mentioned that:
```
Another way is to read all bootstrap peers in the config, and 
if the node's peer ID is the same as one of the bootstrap nodes, it can be marked as a bootstrap node. 
In this case, we keep the config intact.
```

PR #763, opened 2023-10-20
test: protect tests from bad random height

b00f:
## Description

This PR ensures that the random height is always greater than 1000.


amirvalhalla (approved):

amirvalhalla:
@b00f  please fix the deadlock and race condition

amirvalhalla (approved):

PR #765, opened 2023-10-21
fix(consensus): strong termination for the binary agreement

b00f:
## Description

This pull request (PR) implements strong termination for binary agreement. 
Any validator who makes a decision in the MainVote step will broadcast the `decide` message.
Any validator who sees a proper and justified `decide` message can decide on the agreed-upon value and terminate immediately.

This is based on section 5.3.1 of the ABBA spec: "Achieving Stronger Termination"

### DIFF
--- a/consensus/cp.go
+++ b/consensus/cp.go
@@ -313,11 +313,68 @@ func (cp *changeProposer) checkJustMainVote(v *vote.Vote) error {
 	}
 }
 
+func (cp *changeProposer) checkJustDecide(v *vote.Vote) error {
+	err := cp.checkCPValue(v, vote.CPValueZero, vote.CPValueOne)
+	if err != nil {
+		return err
+	}
+	j, ok := v.CPJust().(*vote.JustDecided)
+	if !ok {
+		return invalidJustificationError{
+			JustType: j.Type(),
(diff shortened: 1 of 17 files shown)


consensus/cp.go
```diff
@@ -313,11 +313,68 @@ func (cp *changeProposer) checkJustMainVote(v *vote.Vote) error {
...
+		cp.cpDecided = 1
+		cp.enterNewState(cp.proposeState)
+	} else if cpValue == vote.CPValueZero {
+		roundProposal := cp.log.RoundProposal(cp.round)
+		if roundProposal == nil {
+			cp.queryProposal()
+		}
+		cp.cpDecided = 0
+		cp.enterNewState(cp.prepareState)
 	} else {
```
amirvalhalla, line 378: it's not better to remove `else` condition, if any condition triggered we could use `return` if not we should `panic`

amirvalhalla (changes requested):
@b00f 

consensus/cp.go
b00f: @amirvalhalla fixed.

amirvalhalla (approved):
@b00f 
thanks

PR #771, opened 2023-10-23
feat: add rotate logger support

himmish:
rotate logger support added after certain days

/claim 
#775


b00f:
Read the issue carefully before submitting the changes.
I ma sorry, but this is not we are looking for

image

PR #772, opened 2023-10-23
chore: updating linters

b00f:
## Description

This PR updates the linters and fix some linting issues.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,44 +1,92 @@
 linters:
-  enable-all: true
-  disable:
-    - scopelint # deprecated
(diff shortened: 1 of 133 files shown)


b00f:
Please don't merge it. We may still update version 0.15.1.

kehiy (changes requested):
@b00f Everything looks good, but one additional thing you can do here.
in lint.yml workflow instead of installing golangci lint and running make check, just use golangci lint github action to do this.

themantre (approved):

themantre (approved):

PR #822, opened 2023-11-21
feat(logger): adding rotate log file after days, compress and maxgae for logger config

### LINKED ISSUE #755: Add option to config for rotating logfiles
Old achieve logs are not rotated

image

Add an option in the config.toml like rotate_log_after_days, which gives the user the possibility to set the number of days after the logs files are rotated.

kehiy:
## Description

we have 3 new options for logger config now.
the reason for a lot of compressed backup files was enabling compress by default, not setting max-age and rotate interval for them.
but now, all of them in configurable.

## Related issue(s)

- Fixes #755

### DIFF
--- a/config/example_config.toml
+++ b/config/example_config.toml
@@ -124,6 +124,18 @@
   # Default is `true`.
  ## colorful = true
 
+  # `max_backups` is the maximum number of old log files to retain.
+  # Default is `0`.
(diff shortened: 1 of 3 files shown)


kehiy:
@b00f tests passed.

b00f (approved):

PR #825, opened 2023-11-26
chore: updating version.go adding more detail to Agent function

### LINKED ISSUE #823: Adding more details in agent hello and structured value

We have an **Agent** function in `version.go` that returns some details about the peer (for example `pactus/0.17.0`). We need a structured value and additional details added:

```go
func Agent() string {
	return fmt.Sprintf("pactus/%s", Version())
}
```

We want agent details similar:

```go
func Agent() string {
	return fmt.Sprintf("node=pactus/version=%s/type={GUI or DAEMON}/os=%s/architecture=%s", Version(), runtime.GOOS, runtime.GOARCH)
}
```

In `fmt.Sprintf()` we have many **key=value** and split-ed by **/**

izikdepth:
- Fixes #823


version/version.go
```diff
@@ -13,7 +13,7 @@ const (
 )
 
 func Agent() string {
-	return fmt.Sprintf("pactus/%s", Version())
+	return fmt.Sprintf("node=pactus/version=%s/os=%s/architecture=%s", Version(), runtime.GOOS, runtime.GOARCH)
```
Ja7ad, line 16: We need another key=value about peer is from GUI or Daemon

Ja7ad (changes requested):

version/version.go
kehiy: @Ja7ad I checked the code, there are some places we are calling the Agent function but we don't have access to find out which client is calling the function.

should we pass it from other modules? (for example, adding a ClientType field to www modules, so in response we can use it to pass to the Aget)

b00f:
@izikdepth @Ja7ad @kehiy

This PR is crucial, and we should ensure that we won't need to refactor it in the future. If you're all okay with it, let's schedule an internal meeting to discuss the best approach.


Question from @izikdepth 
Using above parameters: `runtime.GOOS`, `runtime.GOARCH`
Can we know if the agent is GUI or CLI?

Ja7ad:
> @izikdepth @Ja7ad @kehiy
> 
> This PR is crucial, and we should ensure that we won't need to refactor it in the future. If you're all okay with it, let's schedule an internal meeting to discuss the best approach.
> 
> Question from @izikdepth Using above parameters: `runtime.GOOS`, `runtime.GOARCH` Can we know if the agent is GUI or CLI?

We can known with global variable or etc before init node, of course need this global has been set in `cmd`.

kehiy:
> Question from @izikdepth 
> Using above parameters: `runtime.GOOS`, `runtime.GOARCH`
> Can we know if the agent is GUI or CLI?

We can have this meeting tomorrow or now if everyone ok(?).

izikdepth:
tomorrow

> > Question from @izikdepth
> > Using above parameters: `runtime.GOOS`, `runtime.GOARCH`
> > Can we know if the agent is GUI or CLI?
> 
> We can have this meeting tomorrow or now if everyone ok(?).

tomorrow

Ja7ad:
@izikdepth pip updated: https://pips.pactus.org/PIPs/pip-14

kehiy:
Fixed in:
- #841

PR #826, opened 2023-11-26
chore: updating version.go adding more detail to Agent function

### LINKED ISSUE #823: Adding more details in agent hello and structured value

izikdepth:
we can't add client type because we are calling Agent() somewhere we don't have access to client type.

- Fixes #823

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -2,6 +2,7 @@ package version
 
 import (
 	"fmt"
+	"runtime"
 )
 
 // These constants follow the semantic versioning 2.0.0 spec (http://semver.org/)


Ja7ad:
please apply your changes on PR https://github.com/pactus-project/pactus/pull/825

PR #831, opened 2023-12-01
feat: subscribing to libp2p eventbus

b00f:
## Description

By subscribing to the libp2p event bus, we can receive notifications when the node's reachability and supported protocols change.


network/notifee.go
```diff
@@ -26,7 +26,12 @@ type NotifeeService struct {
 func newNotifeeService(ctx context.Context, host lp2phost.Host, eventChannel chan<- Event, peerMgr *peerMgr,
 	protocolID lp2pcore.ProtocolID, bootstrapper bool, log *logger.SubLogger,
 ) *NotifeeService {
-	eventSub, err := host.EventBus().Subscribe(lp2pevent.WildcardSubscription)
+	evts := []interface{}{
```
kehiy, line 29: is that `evt` correct? or should be `event`?

b00f: Done

kehiy (approved):
looks good to me.

PR #837, opened 2023-12-05
feat(consensus): consensus fast path

b00f:
## Description

Implementing fast path for the consensus protocol


Ja7ad:
@b00f We accept https://pips.pactus.org/PIPs/pip-10 for implementation.

PR #843, opened 2023-12-09
feat: implementing pip-15

amirvalhalla:
## Description

This PR implements [PIP-15](https://pips.pactus.org/PIPs/pip-15).


store/account.go
```diff
@@ -1,73 +1,89 @@
...
+const (
+	accLruCacheSize = 1024
+)
+
 type accountStore struct {
-	db         *leveldb.DB
-	addressMap map[crypto.Address]*account.Account
-	total      int32
+	db           *leveldb.DB
+	addrLruCache *lru.Cache[crypto.Address, *account.Account]
```
b00f, line 18: I think we can rename it to `addrCache`

store/store.go
```diff
@@ -43,16 +46,27 @@ func tryGet(db *leveldb.DB, key []byte) ([]byte, error) {
...
+		logger.Trace("database get error", "error", err, "key", key)
 		return nil, err
 	}
 	return data, nil
 }
 
+func tryHas(db *leveldb.DB, key []byte) bool {
+	ok, err := db.Has(key, nil)
+	if err != nil {
+		logger.Error("database has error", "error", err, "key", key)
```
b00f, line 58:
```suggestion
		logger.Error("database `has` error", "error", err, "key", key)
```

store/store.go
```diff
@@ -43,16 +46,27 @@ func tryGet(db *leveldb.DB, key []byte) ([]byte, error) {
 	data, err := db.Get(key, nil)
 	if err != nil {
 		// Probably key doesn't exist in database
-		logger.Trace("database error", "error", err, "key", key)
+		logger.Trace("database get error", "error", err, "key", key)
```
b00f, line 49:
```suggestion
		logger.Trace("database `get` error", "error", err, "key", key)
```

b00f (changes requested):

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/pactus-project/pactus
 go 1.21
 
 require (
+	github.com/eapache/queue/v2 v2.0.0-20230407133247-75960ed334e4
```
b00f, line 6: Please use our LinkedList

node/node.go
```diff
@@ -61,7 +61,8 @@ func NewNode(genDoc *genesis.Genesis, conf *config.Config,
 	txPool := txpool.NewTxPool(conf.TxPool, messageCh)
 
 	// TODO implement dequeue for recent transaction
-	str, err := store.NewStore(conf.Store)
+	str, err := store.NewStore(conf.Store,
+		genDoc.Params().TransactionToLiveInterval, genDoc.Params().SortitionInterval)
```
b00f, line 65: We can pass these parameters through the config as private member.
Similar code can be found here: https://github.com/pactus-project/pactus/blob/a032b7befaef748617c9d9594c7418f36144fda8/sync/config.go#L18

b00f (changes requested):

b00f:
Testing these change is super important. Please make sure all functionalities is tested properly

node/node.go
amirvalhalla: @b00f 
At first, I agreed with you about this change, but you didn't consider that hardcoding takes away flexibility from both users and us?

b00f: We are not going to hardcore anything. Instead of passing new agruments we can use private field in config. Then we get some advantages:
1- We can define parameters with more semantic names in config like: `TxCacheSize` or `SortitionSessCacheSize`
2- We don't need to change the method signature
3- We can assign values based on the some logic outside store, example from `params`

b00f (changes requested):

util/linkedmap/linkedmap.go
```diff
@@ -87,6 +87,13 @@ func (lm *LinkedMap[K, V]) TailNode() *ll.Element[Pair[K, V]] {
 	return ln
 }
 
+func (lm *LinkedMap[K, V]) RemoveTail() {
+	tail := lm.list.Tail
```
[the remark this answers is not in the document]
amirvalhalla: In the `LinkedMap`, there is a method called `Remove ` that takes a key as a parameter and removes the corresponding entry. Could you clarify what you mean by duplicated code? Are you suggesting removing the `RemoveTail ` method and instead using the `Remove(key)` method? Since we already have the `TailNode ` method, we can obtain the key from it and then pass it to the `Remove() ` method to delete the corresponding item.

go.mod
b00f: Remove this line form go.mod

b00f (changes requested):
@amirvalhalla Thanks.
It is almost done, you did amazing Job.

node/node.go
```diff
@@ -154,6 +154,7 @@ func (n *Node) Stop() {
 }
 
 // these methods are using by GUI.
+
```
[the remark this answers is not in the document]
amirvalhalla: I added a new line not because the comment is related to the`ConsManager ` method; in fact, I included it to indicate to others that these methods are for the GUI, as recommended by the Golang Linter.

util/tripleslice/tripleslice.go
```diff
@@ -0,0 +1,81 @@
+package tripleslice
+
+type Triple[K comparable, V any] struct {
+	FirstElement  K
+	SecondElement V
+	ThirdElement  int
```
b00f, line 6: What is the purpose of the third item?

util/tripleslice/tripleslice.go
```diff
@@ -0,0 +1,81 @@
+package tripleslice
+
+type Triple[K comparable, V any] struct {
+	FirstElement  K
```
b00f, line 4: please rename it to first and second

util/tripleslice/tripleslice.go
```diff
@@ -0,0 +1,81 @@
...
+func (ps *TripleSlice[K, V]) Has(index int) bool {
+	if index >= ps.capacity {
+		return false
+	}
+
+	return ps.pairs[index] != nil
+}
+
+// Get returns the Triple at the specified index. If the index doesn't exist, it returns nil.
+func (ps *TripleSlice[K, V]) Get(index int) *Triple[K, V] {
```
b00f, line 48: The internal way we store should be hidden from the user,

b00f (changes requested):

go.mod
```diff
@@ -53,7 +54,6 @@ require (
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-task/slim-sprig v0.0.0-20230315185526-52ccab3ef572 // indirect
 	github.com/godbus/dbus/v5 v5.1.0 // indirect
-	github.com/gofrs/flock v0.8.1 // indirect
```
b00f, line 56: Please revert this file, we use flock as new dependency

store/account.go
```diff
@@ -9,65 +10,83 @@ import (
...
+		return nil, err
+	}
+
+	acc, err = account.FromBytes(rawData)
+	if err != nil {
+		return nil, err
+	}
+
+	as.addrCache.Add(addr, acc)
+	return acc, nil
```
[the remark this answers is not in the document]
b00f: Here, I still feel that we need to review it again. We are returning an account pointer here, and it can be changed, which could potentially manipulate the cache.

store/tx.go
```diff
@@ -17,24 +19,50 @@ type blockRegion struct {
 func txKey(id tx.ID) []byte { return append(txPrefix, id.Bytes()...) }
 
 type txStore struct {
-	db *leveldb.DB
+	db          *leveldb.DB
+	txCache     *linkedmap.LinkedMap[tx.ID, uint32]
```
b00f, line 23: The name is confusing, rename it to "txIDCache"

b00f (changes requested):
Thanks @amirvalhalla. It is Almost done, just small changes before merging.

b00f (approved):

PR #845, opened 2023-12-10
fix: removing BasicCheck for hash

b00f:
## Description

This PR removes `BasicCheck` method from the hash.


amirvalhalla (approved):
@b00f  thanks.

PR #846, opened 2023-12-11
fix(cmd): changing home directory for root users

### LINKED ISSUE #842: Issue regarding initilizing node
## Description

> If you 
> ```./pactus-daemon init``` 
> while you are in normal user you can just launch it with 
> ```./pactus-daemon start``` 
> if you are root user you have to specify with 
> ```./pactus-daemon start -w ~/pactus```. 
> Now a third difference is when people init like this,
> ```./pactus-daemon init  -w ~/pactus --testnet``` 
> it then creates a testnet folder.

This is reported by @Mr-HoDL58 in Discord

b00f:
## Description

This PR set working directory to `/root/pactus` instead of `/pactus` for root users (or inside docker)

It is going to fix #842

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -214,13 +214,10 @@ func PrintJSONObject(obj interface{}) {
 func PactusHomeDir() string {
 	home := ""
 	usr, err := user.Current()
-	if err == nil {
-		// Running as root, probably inside docker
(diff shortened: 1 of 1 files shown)


amirvalhalla (approved):
@b00f thanks

PR #855, opened 2023-12-12
feat: isAlreadyRunning checks on start

### LINKED ISSUE #805: Resource temporarily unavailable

image

jkcs:
## Description

Briefly describe the changes introduced by this pull request.

## Related issue(s)

- Fixes #805
/claim #805

### DIFF
--- a/cmd/daemon/start.go
+++ b/cmd/daemon/start.go
@@ -1,11 +1,17 @@
 package main
 
 import (
+	"bytes"
+	"fmt"
 	"net/http"
 	_ "net/http/pprof" // #nosec
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
 	"time"
 
 	"github.com/pactus-project/pactus/cmd"
@@ -72,11 +78,11 @@ func buildStartCmd(parentCmd *cobra.Command) {
 		cmd.FatalErrorCheck(err)
 
 		// Write current PID to the file
-		err := ioutil.WriteFile(pidFile, []byte(fmt.Sprintf("%d", os.Getpid())), 0666)
+		err = os.WriteFile(pidFile, []byte(fmt.Sprintf("%d", os.Getpid())), 0666)
 		cmd.FatalErrorCheck(err)
 
 		cmd.TrapSignal(func() {
-			os.Remove(pidFilePath)
+			os.Remove(pidFile)
 			node.Stop()
 			cmd.PrintInfoMsgf("Exiting ...")
 		})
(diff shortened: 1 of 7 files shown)


b00f:
@amirvalhalla @Ja7ad @kehiy Please review

b00f:
We need to test it in Linux, Windows  and MacOS

jkcs:
The basic idea is to create a .pactus.pid file in the file directory during runtime to record the process ID (PID) while the program is running. When the program is shut down normally, the .pactus.pid file is deleted. When starting simultaneously, it checks whether there is a .pactus.pid file in the folder and whether the corresponding PID is running to determine if it is a duplicate startup.

cmd/daemon/start.go
```diff
@@ -72,3 +91,35 @@ func buildStartCmd(parentCmd *cobra.Command) {
...
+			return true // PID found and process is running
+		}
+	}
+	return false
+}
+
+// pidExists checks if a given PID is currently active.
+func pidExists(pid int) bool {
+	if runtime.GOOS == "windows" {
+		windowsCmd := exec.Command("tasklist", "/FI", "PID eq "+strconv.Itoa(pid))
```
jkcs, line 109: Due to the lack of a built-in method in Go language to check whether a PID program is running in the Windows system, the built-in tasklist command is used. This approach may have potential issues.

cmd/daemon/start.go
```diff
@@ -72,3 +91,40 @@ func buildStartCmd(parentCmd *cobra.Command) {
...
+
+// pidExists checks if a given PID is currently active.
+func pidExists(pid int) bool {
+	if pid <= 0 {
+		return false
+	}
+
+	if runtime.GOOS == "windows" {
+		pidStr := strconv.Itoa(pid)
+		windowsCmd := exec.Command("tasklist", "/FI", "PID eq "+pidStr)
```
kehiy, line 114:
```suggestion
		windowsCmd := exec.Command("tasklist", "/FI", "PID eq "+pidStr) //#nosec 204
```

you can fix this issue, with disabling this rule for this line only.

kehiy (changes requested):

cmd/daemon/start.go
jkcs: Thank you for the this. 👍🏻

amirvalhalla:
@b00f @jkcs 
Actually, I don't agree with that because, in my opinion, the better way is to obtain the process ID from `syscall`. Then, for checking if the process exists, we could call `os.FindProcess(pid)`. Also, the issue with creating the file is that if the user removes it, we can't detect duplicate processes anymore.

jkcs:
I know that in Windows, we can use `syscall.CreateMutex` for this purpose. However, I'm not sure how to achieve similar functionality in other systems.

amirvalhalla:
> I know that in Windows, we can use `syscall.CreateMutex` for this purpose. However, I'm not sure how to achieve similar functionality in other systems.

I think we could find a way to handle this. @b00f  what is your opinion?

jkcs:
If it is allowed to use `syscall.Open` to open the file in exclusive mode, I can modify it.

b00f:
@jkcs 

We had no choice but to update the base branch, and unfortunately, your PR is now detached from the main branch. I will close this PR; please open a new one. We are so sorry for the inconvenience.

b00f:
@jkcs,

The PR you submitted almost fixed the issue. That is really great! If we can have it in the next release, it would be fantastic. Please let me know if you encounter any difficulties while creating a new PR.

jkcs:
@b00f I want to confirm, is this just a port occupancy issue?

b00f:
@jkcs,

It can be. Running an instance two times will cause issues, either on ports or DB instances. We cannot create two DB instances at the same time.

However, users are currently facing this issue and reporting it on Discord as "Resources temporarily unavailable." Their nodes can't be started because they usually run the node on a VPS using an application like the [screen](https://www.geeksforgeeks.org/screen-command-in-linux-with-examples/) command. They often forget that there is a node running in the background. So, if we can report that Pactus is running, it would be a good hint for them to debug.

I believe Bitcoin also has a way to check if the daemon is running, similar to what you did. It might be worthwhile to check their repository. When I tried to run Bitcoin for the second time, I received this error:

```bash
./bitcoind
Error: Cannot obtain a lock on the data directory /home/mostafa/.bitcoin. Bitcoin Core is probably already running.
```

I hope this information helps.

b00f:
@jkcs 

FYI, here is the bitcoin implementation: 
https://github.com/search?q=repo%3Abitcoin%2Fbitcoin%20LockDirectory&type=code

jkcs:
New PR #871

PR #867, opened 2023-12-13
chore(network): implementing Stop for peer manager

kehiy:


network/peermgr.go
```diff
@@ -70,7 +70,10 @@ func (mgr *peerMgr) Start() {
 
 // Stop stops the Bootstrap.
 func (mgr *peerMgr) Stop() {
-	// TODO: complete me
+	mgr.logger.Info("stopping peer manager")
+	if err := mgr.host.Close(); err != nil {
```
b00f, line 74: Why we need stop host here? 
Did you check [here](https://github.com/pactus-project/pactus/blob/a032b7befaef748617c9d9594c7418f36144fda8/network/network.go#L290C2-L290C2)

kehiy: I thought it was a part of peer manager. so, can we save peer data somewhere in the working directory instead when sopping the peer manager? (we can try to retrieve them in start function)

b00f: Yes, This is a good idea, then once the node starts we have some address to dial. But we need to R&D.
Some nodes are behind NAT. How we know the address is for the public node or not?

Good found @kehiy

PR #871, opened 2023-12-22
feat: check already running by lock file

### LINKED ISSUE #805: Resource temporarily unavailable

jkcs:
## Description

Briefly describe the changes introduced by this pull request.

Referenced this https://github.com/pactus-project/pactus/pull/855#issuecomment-1867283627

## Related issue(s)

- Fixes #805
/claim #805

### DIFF
--- a/cmd/daemon/start.go
+++ b/cmd/daemon/start.go
@@ -46,7 +46,7 @@ func buildStartCmd(parentCmd *cobra.Command) {
 		}
 
 		if !locked {
-			cmd.PrintWarnMsgf("Another instance is already running. lock file '%s'.", lockFilePath)
+			cmd.PrintWarnMsgf("Could not lock '%s', another instance is running?", lockFilePath)
 			return
 		}
 
--- a/cmd/daemon/start.go
+++ b/cmd/daemon/start.go
@@ -80,7 +80,7 @@ func buildStartCmd(parentCmd *cobra.Command) {
 		cmd.FatalErrorCheck(err)
 
 		cmd.TrapSignal(func() {
-			fileLock.Unlock()
+			_ = fileLock.Unlock()
 			node.Stop()
 			cmd.PrintInfoMsgf("Exiting ...")
 		})
--- a/cmd/daemon/start.go
+++ b/cmd/daemon/start.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"time"
 
+	flock "github.com/gofrs/flock"
 	"github.com/pactus-project/pactus/cmd"
 	"github.com/pactus-project/pactus/wallet"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/daemon/start.go
```diff
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"time"
 
+	flock "github.com/gofrs/flock"
```
b00f, line 10: We try to avoid adding any new dependency to the project. 
Any strong reason to add it?

jkcs: Flock handles compatibility issues across different systems and provides a unified API.

amirvalhalla: I think instead of adding a new dependency, you can use sync.RWMutex, which gives us a write locker based on the file system mechanism. For a file, you could only have one writer, and I believe this will solve the issue.

amirvalhalla, line 10: I think instead of adding a new dependency, you can use sync.RWMutex, which gives us a write locker based on the file system mechanism. For a file, you could only have one writer, and I believe this will solve the issue.

amirvalhalla (changes requested):
please resolve the requested changes.

cmd/daemon/start.go
jkcs: Based on my understanding, sync.RWMutex is a lock for synchronization between two threads, and it doesn't have the capability to accomplish this. 
> For a file, you could only have one writer

`flock` is designed to handle such tasks

amirvalhalla: I meant that you can use these solutions, etc., to prevent new dependencies.

b00f: If I understood correctly, `flock` creates a file in read-only mode and locks it [exclusively](https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock.go#L128C52-L128C56). I believe this dependency is not too heavy, so we can include it. Good find, @jkcs!

b00f: The flock is a file-locking system @amirvalhalla . Not process locking

cmd/daemon/start.go
```diff
@@ -34,6 +35,21 @@ func buildStartCmd(parentCmd *cobra.Command) {
 		err := os.Chdir(workingDir)
 		cmd.FatalErrorCheck(err)
 
+		// Define the lock file path
+		lockFilePath := filepath.Join(workingDir, ".pactus.lock")
+		fileLock := flock.New(lockFilePath)
+
+		locked, err := fileLock.TryLock()
```
b00f, line 42: Good code.

cmd/daemon/start.go
```diff
@@ -34,6 +35,21 @@ func buildStartCmd(parentCmd *cobra.Command) {
...
+		if err != nil {
+			// handle unable to attempt to acquire lock
+			cmd.FatalErrorCheck(err)
+		}
+
+		if !locked {
+			cmd.PrintWarnMsgf("Another instance is already running. lock file '%s'.", lockFilePath)
```
b00f, line 49:
```suggestion
			cmd.PrintWarnMsgf("Could not lock '%s', another instance is running?", lockFilePath)
```

b00f (approved):

b00f:
@jkcs Thanks. It is exactly what we were looking for.

amirvalhalla (approved):

amirvalhalla (approved):

themantre:
Please don't merge it till releasing version 0.18.2

b00f:
@kehiy  Please test this PR on Windows
@amirvalhalla Please test this PR on MacOS

b00f:
@jkcs , Thanks for contributing to Pactus. We hope to see more of you on Pactus. The main goal of the bounty program is to find talented developers like you. If you're interested, you can join the Pactus development team. Please join our Discord server, as we have more discussions about core development there: https://discord.gg/H5vZkNnXCu

jkcs:
Thank you for the invitation. I have just joined.

kehiy:
@b00f it's merged, still need to be tested on the windows?

b00f:
>  it's merged, still need to be tested on the windows?
Yes please. Just run node twice.

PR #877, opened 2023-12-27
feat(network): database-backed peer store

### LINKED ISSUE #866: Keeping peer list when node goes off
We need to keep peers somewhere when node goes off, we can use them when node try to start and sync again.

kehiy:
## Description

Adding a new `peers.json` file in the working directory which when the node goes off will keep track of all connected peers. when we start the node again, we have all of the previous nodes to connect in the peer store.

The `peer.json` looks like this:
```json
{
    "peers": {
        "12D3KooWACPieBCFwiF4BfaDAG9EQUk5QgCWvLc3ZYm4LGrW9np5": {
            "multi_address": "/ip4/207.211.157.238/udp/21777/quic-v1",
            "direction": 2
        },
        "12D3KooWAQzz1GtMa6CQRkcbzavaiQbuYPmyKjpmFEjTE9A5Cb8y": {
            "multi_address": "/ip4/146.56.144.37/udp/21777/quic-v1",
            "direction": 2
        },
        "12D3KooWAspuF1ZMc77e9d2efbZt28MPXDmiLj6QkhGbFe7AA1Cs": {
            "multi_address": "/ip4/168.138.184.213/udp/21777/quic-v1",
            "direction": 2
        },
        "12D3KooWBGNEH8NqdK1UddSnPV1yRHGLYpaQUcnujC24s7YNWPiq": {
            "multi_address": "/ip4/13.115.190.71/tcp/21777",
            "direction": 2
        }
    }
}
```
> NOTE: most of the time it's around 30 peers, in this example, I showed 4 of them only.

The node owner can change the peer store path in config.

## Related issue(s)

- Fixes #866


b00f:
Two general comments:

1. The nodes behind NAT also publish a public address. This should be considered.
2. There is no need to add a new item to the configuration.

network/peermgr.go
```diff
@@ -157,3 +181,39 @@ func (mgr *peerMgr) CheckConnectivity() {
 		}
 	}
 }
+
+type PeerStore struct {
```
b00f, line 185: Why we can't use peerMgr?

kehiy:
> Two general comments:
> 
> 1. The nodes behind NAT also publish a public address. This should be considered.
> 2. There is no need to add a new item to the configuration.

for second issue, should I use `peers.json` as a constant?

network/peermgr.go
kehiy: Peer ID's will be saved in Unicode format and make problems.

kehiy: This `lp2ppeer.ID` type makes a problem for us. Peer IDs going to be saved as the unicodes and then we can't retrive them at all. we can use `map[string]PeerInfo` without defineing it as a type ot struct. but the main problem is json tags.

b00f:
We need more R&D. The libp2p I think has persistent store for peers.

kehiy:
> We need more R&D. The libp2p I think has persistent store for peers.

The persistent store for peers is now deprecated.
see:
- https://github.com/libp2p/go-libp2p/issues/2329

So, we can implement it ourselves. just we have 2 issues, the peers with private reachability and having a new struct for peer store (as I said we can't use current map on peer manager).

About private peers, I thought if we can have reachability field on peerInfo struct too.

b00f:
Why did you close it?

PR #898, opened 2024-01-02
feat(gRPC): defining network and peers info response's properly

### LINKED ISSUE #425: grpc: `GetPeerInfo` and `GetNetworkInfo` are not well defined
We try our best to define gRPC APIs as much as possible close to the Bitcoin JSON-RPC APIs. It looks the `GetNetworkInfo` and `GetPeerInfo` in our proto file are not well defined. 

## GetNetworkInfo 
In Bitcoin, [getnetworkinfo](https://developer.bitcoin.org/reference/rpc/getnetworkinfo.html) is designed to "Returns an object containing various state info regarding P2P networking". 
For example, here is the result of calling `getnetworkinfo` in my local machine.
```json
{
  "version": 220000,
  "subversion": "/Satoshi:22.0.0/",
  "protocolversion": 70016,
  "localservices": "0000000000000408",
  "localservicesnames": [
    "WITNESS",
    "NETWORK_LIMITED"
  ],
  "localrelay": true,
  "timeoffset": 0,
  "networkactive": true,
  "connections": 3,
  "connections_in": 0,
  "connections_out": 3,
  "networks": [
    {
      "name": "ipv4",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "ipv6",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "onion",
      "limited": true,
      "reachable": false,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "i2p",
      "limited": true,
      "reachable": false,
      "proxy": "",
      "proxy_randomize_credentials": false
    }
  ],
  "relayfee": 0.00001000,
  "incrementalfee": 0.00001000,
  "localaddresses": [
    {
      "address": "....",
      "port": 8333,
      "score": 1
    },
    {
      "address": "....",
      "port": 8333,
      "score": 1
    }
  ],
  "warnings": ""
}
```

[GetNetworkInfo](https://github.com/pactus-project/pactus/blob/6c575f16f3335268784a17b46d24897b6b15b42e/www/grpc/proto/network.proto#L8) in our proto returns the list of peers info. It is not compatible with the `getnetworkinfo`. So we need to redefine the `GetNetworkInfoResponse` properly. 
We should be able to get information about the network from libp2p. We can do it in another PR, but it should be accessible in this gRPC call.

## GetPeerInfo
In Bitcoin, [getpeerinfo](https://developer.bitcoin.org/reference/rpc/getpeerinfo.html) is designed to "Returns data about each connected network node as a json array of objects.". 
For example, here is the result of calling `getpeerinfo` in my local machine.

```json
[
  {
    "id": 0,
    "addr": "69.228.134.38:8333",
    "addrbind": "192.168.1.203:50606",
    "addrlocal": "115.134.6.105:50606",
    "network": "ipv4",
    "services": "0000000000000409",
    "servicesnames": [
      "NETWORK",
      "WITNESS",
      "NETWORK_LIMITED"
    ],
    "relaytxes": false,
    "lastsend": 1668569765,
    "lastrecv": 1668569765,
    "last_transaction": 0,
    "last_block": 1668569765,
    "bytessent": 9082,
    "bytesrecv": 118829731,
    "conntime": 1668569714,
    "timeoffset": -2,
    "pingtime": 0.253437,
    "minping": 0.253437,
    "version": 70016,
    "subver": "/Satoshi:23.0.0/",
    "inbound": false,
    "bip152_hb_to": false,
    "bip152_hb_from": false,
    "startingheight": 763371,
    "synced_headers": 763371,
    "synced_blocks": 512398,
    "inflight": [
      512566,
      512569,
      512570,
      512571,
      512574,
      512576,
      512579,
      512581,
      512582,
      512584,
      512587,
      512588,
      512591,
      512593,
      512611,
      512614
    ],
    "addr_processed": 0,
    "addr_rate_limited": 0,
    "permissions": [
    ],
    "minfeefilter": 0.00000000,
    "bytessent_per_msg": {
      "getdata": 7677,
      "getheaders": 1053,
      "ping": 32,
      "pong": 32,
      "sendaddrv2": 24,
      "sendcmpct": 66,
      "sendheaders": 24,
      "verack": 24,
      "version": 126,
      "wtxidrelay": 24
    },
    "bytesrecv_per_msg": {
      "block": 118362994,
      "feefilter": 32,
      "getheaders": 1053,
      "headers": 106,
      "ping": 32,
      "pong": 32,
      "sendaddrv2": 24,
      "sendcmpct": 66,
      "sendheaders": 24,
      "verack": 24,
      "version": 126,
      "wtxidrelay": 24
    },
    "connection_type": "block-relay-only"
  },
  {
    "id": 1,
    ...
  },
  {
    "id": 4,
    ...
  },
  {
    "id": 5,
    ...
  },
  {
    "id": 7,
   ...
  }
]
```

[GetPeerInfo](https://github.com/pactus-project/pactus/blob/6c575f16f3335268784a17b46d24897b6b15b42e/www/grpc/proto/network.proto#L9) in our proto returns our local node info. It is not compatible with the `getpeerinfo`. So we need to redefine the `GetPeerInfoResponse` like:

```proto
message GetPeerInfoResponse {
  repeated PeerInfo peers = 1;
}

```

 It should return the list of connected peers to our local node. The [`PeerInfo`](https://github.com/pactus-project/pactus/blob/6c575f16f3335268784a17b46d24897b6b15b42e/www/grpc/proto/network.proto#L25) struct can be used here, but not for `GetNetworkInfo`. They might have some common fields like version, but they are not the same.

kehiy:
## Description

With this PR we have our network APIs more well defined. I defined a Network name also as a network API.

Our gRPC methods for network have we have (network and node info), return more detaild responses and they contain data releated to themselves only.

Example response for network info:
```json
{
  "totalSentBytes": 9405,
  "totalReceivedBytes": 1920935,
  "networkName": "pactus-testnet-v2",
  "connectedPeersCount": 27,
  "connectedPeers": [
    {
      "status": 3,
      "moniker": "",
      "agent": "node=pdaemon/node-version=v0.18.4/protocol-version=1/os=linux/arch=arm64",
      "peerId": "ACQIARIg0JSiRevvYC9519IiKV4079EX1JqFPdkMcGtbU7drEJc=",
      "consensusKeys": [
        "tpublic1pjmngf884hnt023sfg2qr0289ytx935jyh70ehr3c2yexcd2haentsh9qz0tk0e6hn46fax98nu50c965syf525u0xsw2uzunew6z8lvjqfkapsqgc7zqmvfdm53twsrsusjady3w0duq0fercrqehpe3avfg2csj"
      ],
      "services": 1,
      "lastBlockHash": "pImH+5u0KnxjhirKjG1cl2JlOiZLPTJ5PK08Hd9djKM=",
      "height": 625793,
      "receivedMessages": 2,
      "invalidMessages": 0,
      "lastSent": "1704287429",
      "lastReceived": "1704287430",
      "sentBytes": {
        "1": "398",
        "2": "22"
      },
      "receivedBytes": {
        "1": "373",
        "2": "26"
      },
      "address": "/ip4/129.153.54.255/udp/21777/quic-v1",
      "direction": "Outbound",
      "protocols": [
        "/floodsub/1.0.0",
        "/ipfs/id/1.0.0",
        "/ipfs/id/push/1.0.0",
        "/libp2p/circuit/relay/0.2.0/stop",
        "/libp2p/dcutr",
        "/pactus-testnet-v2/gossip/v1/kad/1.0.0",
        "/pactus-testnet-v2/stream/v1"
      ],
      "totalSessions": 0,
      "completedSessions": 0
    }
]
}
```

Node info:
```json
{
  "moniker": "",
  "agent": "node=pactus-daemon/node-version=v0.19.0-beta/protocol-version=1/os=linux/arch=amd64",
  "peerId": "ACQIARIgEkKb9wqNNqkQxi3xLXFOOzvQRtlGS+Yr5NTjpOI0uj8=",
  "startedAt": "1704287428",
  "reachability": "Private",
  "services": [
    1
  ],
  "servicesNames": [
    "NETWORK"
  ],
  "addrs": [],
  "protocols": [
    "/libp2p/circuit/relay/0.2.0/stop",
    "/ipfs/id/1.0.0",
    "/ipfs/id/push/1.0.0",
    "/pactus-testnet-v2/stream/v1",
    "/floodsub/1.0.0"
  ]
}
```

## Related issue(s)

- Fixes #425


network/interface.go
```diff
@@ -124,4 +124,7 @@ type Network interface {
 	NumConnectedPeers() int
 	ReachabilityStatus() string
 	HostAddrs() []string
+	Name() string
```
themantre, line 127: Name is always fix and won't change. I am not sure why you make an API for that.

network/interface.go
```diff
@@ -124,4 +124,7 @@ type Network interface {
...
+	DHTSize() int32
```
themantre, line 128: What is the reason for user to know the DHT size?

network/interface.go
```diff
@@ -124,4 +124,7 @@ type Network interface {
...
+	Protocols() []string
```
themantre, line 129: Supported protocols can get from other peers.

themantre (commented):
I think this PR doesn't help much. Explain why we need to add these 3 more APIs to the network?

network/interface.go
kehiy: I considered if someone run a private/local network with pactus protocol, so probably they will have a different name. also a fork of pactus can have a different name for the network.

network/interface.go
kehiy: Do you mean the `GetPeersInfo` method?

network/interface.go
b00f: Hoom, not really a good reason to add another API.

network/interface.go
b00f: How about other information?
In this level, user should look at the log, if it is important we can add a log for DHT size.

network/interface.go
b00f: Protocol negotiation. I don't think LibP2P returns all supported protocols by a node.

kehiy: Protocol negotiation works on Pactus network, this function is used for network info in gRPC and HTTP modules.

network/interface.go
kehiy: yes, we can add it to logs or metrics also. 
about other information, these are was all of infos we currently have.
the purpose of this PR is definig peers info and network info sparated and better. about the information itself, we need to work on it in another PR, because there is more data that we can extract from network.

network/interface.go
kehiy: we can remove it. but I belive having network name can help. also specifing the network with chain ID or something like that can help in future. this kind of stuff helps us in different golas in future.

b00f: This is fixed per chain. We can even get it from the config. Check config [here](https://github.com/pactus-project/pactus/blob/440761955621b04f7a6eb484fe9c3dd1d03065d8/config/config.go#L121C2-L121C2)

network/interface.go
b00f: Some information is useful inside the module, but not helpful to be exported. They are designed for the internal usage, like DHT size. 

We shouldn't add new API unless there's a really good reason for it.

network/interface.go
b00f: How do you want to extract the protocol? The current implementation is wrong. It just return stream protocol id two times: `/pactus-testnet-v2/stream/v1, /pactus-testnet-v2/stream/v1`

The supported protocols should get from LibP2p2, like:
`/floodsub/1.0.0,/ipfs/id/1.0.0,/ipfs/id/push/1.0.0,/libp2p/circuit/relay/0.2.0/stop,/libp2p/dcutr,/meshsub/1.0.0,/meshsub/1.1.0,/pactus-testnet-v2/stream/v1`

network/interface.go
kehiy: I agree, we can remove the DHT size.

network/interface.go
kehiy: yeap, current approach is not fine. let me update that.

network/interface.go
kehiy: yes, I knew we can change it in config per chain also. this function is just a API to show this in network info method, because it's  **Network** info. so I think we should also get the network name from it. for example a same SDK can call different pactus networks (testnet/mainnet) and also different pactus forks, and also detect which network it's calling with this parameter.

network/interface.go
kehiy: fixed.

network/interface.go
b00f: Ok, I am convinced. Let's add `Name` API to the network module.

network/network.go
```diff
@@ -398,3 +398,15 @@ func (n *network) HostAddrs() []string {
...
 }
+
+func (n *network) Name() string {
+	return n.config.NetworkName
+}
+
+func (n *network) Protocols() []string {
+	protocols := []string{}
+	for _, p := range n.host.Mux().Protocols() {
+		protocols = append(protocols, string(p))
```
b00f, line 409: Please cover it by test.

version/version.go
```diff
@@ -13,12 +13,12 @@ const (
 	minor           uint   = 19
 	patch           uint   = 0
 	meta            string = "beta"
-	protocolVersion uint   = 1
+	ProtocolVersion uint   = 1
```
b00f, line 16: Keep it private and add a public function name it `ProtocolVersion()`

www/grpc/proto/network.proto
```diff
@@ -7,17 +7,18 @@ option java_package = "pactus.network";
...
 message GetNetworkInfoRequest {}
 
 message GetNetworkInfoResponse {
-  int32 total_sent_bytes = 1;
-  int32 total_received_bytes = 2;
-  int64 started_at = 3;
-  repeated PeerInfo peers = 4;
-  map<int32, int64> sent_bytes = 5;
-  map<int32, int64> received_bytes = 6;
+  int32 protocol_version = 1;
```
b00f, line 16: It is better to add `protocol_version` to the  `GetNodeInfoResponse`

www/grpc/proto/network.proto
```diff
@@ -7,17 +7,18 @@ option java_package = "pactus.network";
...
+  int32 connected_peers = 2;
+  string network_name = 3;
```
b00f, line 18: Make it number 1.

www/http/nrtwork.go
```diff
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"fmt"
 	"net/http"
 	"time"
```
b00f, line 7: Please rename the file to `network.go`

www/grpc/proto/network.proto
kehiy: Also, the protocol version is available in agent in node info.

network/network_test.go
```diff
@@ -251,6 +251,33 @@ func TestNetwork(t *testing.T) {
 		assert.Equal(t, networkX.ReachabilityStatus(), "Private")
 	})
 
+	t.Run("Supported Protocols", func(t *testing.T) {
+		protosNetB := networkB.Protocols()
```
b00f, line 255: Kay, keep only one, for example Bootstrap one.

kehiy: Fixed.

sync/bundle/message/hello.go
```diff
@@ -26,11 +26,11 @@ type HelloMessage struct {
 }
 
 func NewHelloMessage(pid peer.ID, moniker string,
-	height uint32, services service.Services, blockHash, genesisHash hash.Hash,
+	height uint32, services service.Services, blockHash, genesisHash hash.Hash, reachability string,
 ) *HelloMessage {
 	return &HelloMessage{
 		PeerID:          pid,
-		Agent:           version.Agent(),
+		Agent:           fmt.Sprintf("%v/%v", version.Agent(), reachability),
```
b00f, line 33:
```suggestion
		Agent:           fmt.Sprintf("%v/reachability=%v", version.Agent(), reachability),
```

sync/handler_hello_test.go
```diff
@@ -99,7 +99,7 @@ func TestParsingHelloMessages(t *testing.T) {
 
 			pub := valKey.PublicKey()
 			assert.Equal(t, p.Status, peerset.StatusCodeKnown)
-			assert.Equal(t, p.Agent, version.Agent())
+			assert.Equal(t, p.Agent, version.Agent()+"/public")
```
b00f, line 102:
```suggestion
			assert.Equal(t, p.Agent, version.Agent()+"/reachability=public")
```

b00f (approved):

Issue #869, opened 2023-12-21
Second Bond transaction with same height will be rejected


b00f:
# Description

If a user sends a Bond transaction, and it is rejected for any reason (such as insufficient funds), the next Bond transaction will always be rejected. The reason for this is that the sandbox has created a transient validator, and therefore, the Bond execution believes that the [public key has been set beforehand](https://github.com/pactus-project/pactus/blob/e6657343499407cea45b9afbae5200399dfa5ca7/execution/executor/bond.go#L35).

Here is the log from the Discord Bot that simulates the situation:

```
faucet-referral tpc1pufckvy8yt6ck95rn09dewl0f56zjjw9ct336ac 899332
error while sending the bond transaction: %!w(string=rpc error: code = Canceled desc = couldn't add to transaction pool: insufficient funds)

faucet-referral tpc1pufckvy8yt6ck95rn09dewl0f56zjjw9ct336ac 822121
public key: public key set)

faucet tpc1pufckvy8yt6ck95rn09dewl0f56zjjw9ct336ac
error while sending the bond transaction: %!w(string=rpc error: code = Canceled desc = couldn't add to transaction pool: invalid public key: public key set)
```

### How To Reproduce

The above log shows how to reproduce the issue.

### What were you expecting?

We can fix it by checking the transient validator to ensure it does not have the same height as the sandbox.

PR #901, opened 2024-01-03
fix(executor): not rejecting bond transaction for bootstrap validator

### LINKED ISSUE #900: Unable to sync the blockchain
## Description

Unable to sync the blockchain in Testnet at block 35861. Restarting the node won't fix the issue.

```
WRN committing block failed, removing block from the cache error="invalid public key: public key is not set" _sync="{☍ 30 ⛃ 240}" height=35861
```

### What Happened

Fixing #869 has backfired. Block 35861 contains a Bond transaction for a bootstrap validator. The bootstrap validator has a public key, but at the same time, the `LastBondingHeight` for the bootstrap validator has been set to zero.

We can fix this by adding a new method to the Sandbox to determine if the Validator is sandboxed or not. However, this will add more complexity to the code. The issue #869 is rare and occurs only once in Testnet. This issue is not permanent; for example, after one day (once the first transaction expires), it will be fixed, or users can broadcast the transaction to other nodes. Fixing both issues will increase the code complexity and make the code harder to understand.

It's better to revert the fix for #869.

b00f:
## Description

This PR reverts PR #870  and fixes issue #900.

## Related issue(s)

- Fixes #900 (Related to #869 )

### DIFF
--- a/sandbox/mock.go
+++ b/sandbox/mock.go
@@ -63,7 +63,7 @@ func (m *MockSandbox) Account(addr crypto.Address) *account.Account {
 	return acc
 }
 
-func (m *MockSandbox) MakeNewAccount(addr crypto.Address) *account.Account {
+func (m *MockSandbox) MakeNewAccount(_ crypto.Address) *account.Account {
 	acc := account.NewAccount(m.TestStore.TotalAccounts())
 	return acc
 }
--- a/sandbox/mock.go
+++ b/sandbox/mock.go
@@ -26,8 +26,6 @@ type MockSandbox struct {
 	TestJoinedValidators map[crypto.Address]bool
 	TestCommittedTrxs    map[tx.ID]*tx.Tx
 	TestPowerDelta       int64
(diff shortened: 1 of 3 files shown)


sandbox/mock.go
```diff
@@ -89,9 +81,6 @@ func (m *MockSandbox) AnyRecentTransaction(txID tx.ID) bool {
 
 func (m *MockSandbox) Validator(addr crypto.Address) *validator.Validator {
 	val, _ := m.TestStore.Validator(addr)
-	if val == nil {
-		return m.TestSandboxVals[addr]
-	}
 	return val
```
kehiy, line 84: what if the validator was nil here? are we check that in the upper layers?

b00f: It can be `nil`. In that case, the new validator should be created using `MakeNewValidator`. Please note that this is a mock and is used only for testing purposes.

kehiy: Ok, got it.

kehiy (approved):
everything looks good.

PR #931, opened 2024-01-10
feat: implement relay service

amirvalhalla:
## Description

implement auto relay and relay service


kehiy (approved):

network/notifee.go
```diff
@@ -84,6 +84,10 @@ func (s *NotifeeService) Stop() {
 	s.lp2pEventSub.Close()
 }
 
+func (s *NotifeeService) Reachability() lp2pnetwork.Reachability {
```
kehiy, line 87: also here you can consider RWlock to prevent this deadlock issue.

network/notifee.go
```diff
@@ -84,6 +84,10 @@ func (s *NotifeeService) Stop() {
...
+	return s.reachability
```
kehiy, line 88: consider adding an RLock here to avoid the deadlock issue.

network/network.go
```diff
@@ -259,6 +268,44 @@ func newNetwork(conf *Config, log *logger.SubLogger, opts []lp2p.Option) (*netwo
 	return n, nil
 }
 
+func findRelayPeers(h func() lp2phost.Host, log *logger.SubLogger) func(ctx context.Context,
+	num int) <-chan lp2ppeer.AddrInfo {
+	return func(ctx context.Context, num int) <-chan lp2ppeer.AddrInfo {
+		log.Debug("try to find relay peers", "num", num)
```
kehiy, line 274: I think it's better to use the `Info` level for logging here. what do you think?

kehiy (changes requested):

network/notifee.go
b00f: No need, This Will call only be the external API and only change once.

b00f (approved):

network/notifee.go
b00f: The Readability is set only once, and the external API calls it. It is not necessary to add a Lock here. We have a test failure. Let's fix that.

kehiy: So, why there is a deadlock issue?

b00f (approved):

PR #943, opened 2024-01-17
chore(daemon): removing password, entropy and validator number as prompt, and adding flags instead

kehiy:
## Description

In this PR I removed the validator number, password, and entropy as prompt and hard-coded values. now we get them as flags from the user, this helps other interfaces and clients like GUI use this `init` command with ease.

### DIFF
--- a/cmd/daemon/init.go
+++ b/cmd/daemon/init.go
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// buildInitCmd builds a sub-command to initialized the Pactus blockchain node.
+// buildInitCmd builds a sub-command to initialize the Pactus blockchain node.
 func buildInitCmd(parentCmd *cobra.Command) {
 	initCmd := &cobra.Command{
 		Use:   "init",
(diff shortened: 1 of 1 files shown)


cmd/daemon/init.go
```diff
@@ -53,15 +60,12 @@ func buildInitCmd(parentCmd *cobra.Command) {
...
-		cmd.PrintLine()
-		cmd.PrintInfoMsgf("Enter a password for wallet")
-		password := cmd.PromptPassword("Password", true)
 
-		cmd.PrintLine()
-		cmd.PrintInfoMsgBoldf("How many validators do you want to create?")
-		cmd.PrintInfoMsgf("Each node can run up to 32 validators, and each validator can hold up to 1000 staked coins.")
-		cmd.PrintInfoMsgf("You can define validators based on the amount of coins you want to stake.")
-		numValidators := cmd.PromptInputWithRange("Number of Validators", 7, 1, 32)
+		if *valNumOpt < 1 || *valNumOpt > 32 {
```
b00f, line 64: We can't change the previous behavior.
I suggest:
1- Don't set default in flags
2- If it is not set (or zero) ask user to set it manually

b00f (changes requested):

cmd/daemon/init.go
```diff
@@ -53,15 +59,12 @@ func buildInitCmd(parentCmd *cobra.Command) {
 			err := wallet.CheckMnemonic(*restoreOpt)
 			cmd.FatalErrorCheck(err)
 		}
-		cmd.PrintLine()
-		cmd.PrintInfoMsgf("Enter a password for wallet")
-		password := cmd.PromptPassword("Password", true)
 
-		cmd.PrintLine()
```
b00f, line 60: Still we break the previous behaviour. If user didn't set the number of validator, it should work as before.

cmd/daemon/init.go
```diff
@@ -53,15 +59,12 @@ func buildInitCmd(parentCmd *cobra.Command) {
...
-		cmd.PrintInfoMsgBoldf("How many validators do you want to create?")
-		cmd.PrintInfoMsgf("Each node can run up to 32 validators, and each validator can hold up to 1000 staked coins.")
-		cmd.PrintInfoMsgf("You can define validators based on the amount of coins you want to stake.")
-		numValidators := cmd.PromptInputWithRange("Number of Validators", 7, 1, 32)
```
b00f, line 64: Still we break the previous behaviour. If user didn't set the number of validator, it should work as before.

b00f: It should be like:
```
var valNum int
if *valNumOpt == 0 {
// .... old code
} else {
// ... new code
}
```

kehiy:
@b00f fixed.

b00f (approved):

PR #945, opened 2024-01-18
feat(grpc): sign transaction using wallet client

b00f:
## Description

This PR adds a signing raw transaction function to the wallet gRPC server. 
This enhancement enables merchants to use the Pactus node for issuing and signing transactions.


amirvalhalla (approved):
@b00f thanks

PR #947, opened 2024-01-19
feat: pactus gui lock support

### LINKED ISSUE #805: Resource temporarily unavailable

jkcs:
## Description

CMD PR #871

check gui already running by lock file

## Related issue(s)

- Fixes #805


b00f (approved):

b00f:
Thanks @jkcs

PR #950, opened 2024-01-19
fix(daemon): keeping previous behavior for password flag, linting CLI messages

kehiy:
## Description

This PR will update the password option in the daemon CLI init subcommand. also adds new rules for CLI and other modules helper messages and descriptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,13 @@ The following commands are available in the Makefile:
 - `make test` runs the tests to ensure that all functionality is working as intended.
 - `make proto` regenerates the corresponding code if you have made any changes to the proto buffer files.
 
+Help and description messages (example: CLI subcommands descriptions) MUST follow this style:
+
+- All messages start with lowercase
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -27,6 +27,13 @@ The following commands are available in the Makefile:
 - `make test` runs the tests to ensure that all functionality is working as intended.
 - `make proto` regenerates the corresponding code if you have made any changes to the proto buffer files.
 
+Help and description messages (example: CLI subcommands descriptions) MUST follow this style:
```
b00f, line 30: I suggest:

```
## Guidelines for CLI Help Messages:

- Start all messages with a lowercase letter.
- Avoid stating defaults in the help string, as Cobra automatically adds them.
- Include a range for flags that accept a range of values.
- Exclude parentheses "()" from help messages.
```

kehiy: I agree, a better way to write that. commited.

b00f (approved):

PR #964, opened 2024-01-26
feat: add bootstrap.json and load in config on build

Ja7ad:


config/config.go
```diff
@@ -79,12 +92,14 @@ func defaultConfig() *Config {
...
 func DefaultConfigMainnet() *Config {
 	conf := defaultConfig()
-	conf.Network.DefaultBootstrapAddrStrings = []string{
-		"/dns/bootstrap1.pactus.org/tcp/21888/p2p/12D3KooWMnDsu8TDTk2VV8uD8zsNSB6eUkqtQs6ttg4bHq9zNaBe",
-		"/dns/bootstrap2.pactus.org/tcp/21888/p2p/12D3KooWM39ag7ghta49qybf7McADgT8FLakTYkCsiBvwdnjuG5q",
-		"/dns/bootstrap3.pactus.org/tcp/21888/p2p/12D3KooWBCPSZWheet6tMoHbVBCDfBwQm4yzCwcQ8hJ6NMCN97sj",
-		"/dns/bootstrap4.pactus.org/tcp/21888/p2p/12D3KooWKg6aLa77yAaqMCb45aH5iQuTr5GzRUWUCJ1sZYB5vnoL",
+
+	bootstrapNodes := make([]Bootstrap, 0)
+	if err := json.Unmarshal(bootstrapNodesBytes, &bootstrapNodes); err == nil {
```
b00f, line 97: Please return err if there is any error here.
Otherwise we have empty list for Bootstrap without any error

config/config.go
```diff
@@ -36,6 +42,13 @@ type Config struct {
 	Nanomsg   *nanomsg.Config   `toml:"nanomsg"`
 }
 
+type Bootstrap struct {
```
b00f, line 45: Rename it to `BootstrapInfo`

config/config.go
```diff
@@ -20,8 +21,13 @@ import (
...
 )
 
-//go:embed example_config.toml
-var exampleConfigBytes []byte
+var (
+	//go:embed example_config.toml
+	exampleConfigBytes []byte
+
+	//go:embed bootstrap.json
+	bootstrapNodesBytes []byte
```
b00f, line 29: bootstrapInfosBytes

config/config.go
kehiy, line 97: Since we are adding JSON file data here to the default config, `TestExampleConfig` will fail.

maybe we can add it to the example config too or add them in another function(?)

Ja7ad: This test have issue on Equal, mainnet config never equal with example

Ja7ad: We handle reverse err == nil and never panic happen here

b00f (approved):

amirvalhalla (approved):

kehiy (approved):

PR #970, opened 2024-01-27
fix: localnet wallet issue

b00f:
## Description

This PR fixes the lcoalnet wallet issue that couldn't query the local node.


kehiy (commented):
This should fix our local-net wallet issue, but we can test it (and improve the wallet test also) by adding a genesis type to the `wallet_test` setup function.

Also, we have some unrelated(?) failing test on network gossip.

b00f:
> This should fix our local-net wallet issue, but we can test it (and improve the wallet test also) by adding a genesis type to the `wallet_test` setup function.
> 
> Also, we have some [unrelated](https://github.com/pactus-project/pactus/actions/runs/7680835411/job/20933348810?pr=970#step:7:224)(?) failing tests on network gossip.

Yes, we have time-to-time issues that came from Libp2p. Ignore them

kehiy (changes requested):
You can fix failing tests by updating the default local net config test to this:

```go
assert.Equal(t, conf.GRPC.Listen, "[::]:50052")
assert.Equal(t, conf.GRPC.Gateway.Listen, "[::]:0")
assert.Equal(t, conf.HTTP.Listen, "[::]:0")
assert.Equal(t, conf.Nanomsg.Listen, "tcp://[::]:40799")
```

kehiy (approved):
Everything looks good to me.
Thanks @b00f.

PR #1044, opened 2024-01-31
chore: add bootstrap address for BlockSync

blocksync247:
## Description

Adding  bootstrap address for BlockSync


b00f:
@blocksync247 Thanks

PR #1053, opened 2024-01-31
chore: add bootstrap address for Tetuani

ETHSev:
## Description

Adding bootstrap address for Tetuani


b00f:
@ETHSev Enter valid address

Issue #545, opened 2023-06-27
Supporting snap for Linux


b00f:
## Describe the problem you'd like to have solved

Some users are having difficulties running Pactus on Linux, especially on Ubuntu 20.04. 
This is because GLIBC_2.32 (intended for Ubuntu 22.04) and higher cannot be installed on Ubuntu 20.04.

### Describe the ideal solution

We should consider packaging Pactus using Snap. 
This approach will enable users with different distros to run Pactus without worrying about missing dependencies.

### Alternatives and current work-around

The only workaround currently available is to compile Pactus directly on the target machine.

Ja7ad:
@b00f we need to support deb and rpm too.

b00f:
Here you can find an example of Bitcoin packaging using snap: https://github.com/bitcoin-core/packaging

laosiji-io:
This can only be compiled by the user themselves.
Don't worry too much, we just need to tell them how to compile it.

b00f:
@laosiji-io I am afraid to say that we are worried. We want to make it easy for users to run Pactus. 
We can't simply say "go and compile." 
We should support a variety of operating systems and packaging.

laosiji-io:
I am currently trying to use SnapCraft to create installation packages.
But there are also some issues that cannot load dependencies like GLIBC_2.29.
Different systems encounter different isssues as well.

b00f:
I am confident that we can overcome these difficulties.

laosiji-io:
At present, the snap installation package can be successfully built locally.
The next step is to publish to the Snap Store and implement automatic build.

kehiy:
Any updates on this PR? I can see other projects providing their releases on these repos but they mentioned that this is a community version and unofficial, such as IPFS Kubo implementation. All repos have them but from the community.

What do you think?

b00f:
I believe AppImage is sufficient for now, as it works with most Linux distributions. There's no need for Snap at this moment.

PR #1068, opened 2024-02-04
add snapcraft.yaml

laosiji-io:
Fixing #545


b00f:
This should be under `.github/releaser/linux` folder.
We build snapcraft only on releasing the software.

snap/snapcraf-all.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - wget
+      - dpkg-dev
+      - libgtk-3-0
+      - libgtk-3-bin
+      - libgtk-3-common
+      - libgtk-3-dev
+    override-pull: |
+      snapcraftctl pull
+    override-build: |
+      PACT_VERSION="1.0.0"
```
b00f, line 65: Do we need to build it again or maybe we can use the current binaries?

b00f:
@laosiji-io any update on this PR?

PR #1071, opened 2024-02-04
feat(network): peer store to keep peers list when node goes off

### LINKED ISSUE #866: Keeping peer list when node goes off

kehiy:
## Description

The Peer store will save all connected peers when the node goes off. when a node tries to start again the `CheckConnectivity` function will loop over peers beside bootstrap nodes. Thus, the chance of getting synced and stable in network connection increases.

## Related issue(s)

- Fixes #866


network/notifee.go
```diff
@@ -92,7 +93,8 @@ func (s *NotifeeService) Connected(_ lp2pnetwork.Network, conn lp2pnetwork.Conn)
 	pid := conn.RemotePeer()
 	s.logger.Info("connected to peer", "pid", pid, "direction", conn.Stat().Direction, "addr", conn.RemoteMultiaddr())
 
-	s.peerMgr.AddPeer(pid, conn.RemoteMultiaddr(), conn.Stat().Direction)
+	ai, _ := lp2ppeer.AddrInfoFromP2pAddr(conn.RemoteMultiaddr())
+	s.peerMgr.AddPeer(pid, *ai, conn.Stat().Direction)
```
b00f, line 97: I think we should not add any peer here, indeed let's add when the `EvtPeerIdentificationCompleted` raised.
Then `PeerMgr` can decided, based on the protocols to save the peer info

b00f: Consider some Bot or junk node may connected to our node

themantre:
@kehiy I noticed LibP2P has a PeerStore. Look [here](https://github.com/libp2p/go-libp2p/blob/e3b0a86e863ff6694c4c7bd052c615cbe135c8c1/defaults.go#L61)

Maybe we can have a persistent version of PeerStore

kehiy:
@themantre about the libp2p peer store we discussed before. that's depreacated and not suggested to use and we decided to don't use it.

PR #1081, opened 2024-02-08
fix: stderr logger in windows os

### LINKED ISSUE #1077: WIndows GUI doesn't create the log file
## Description

The Windows GUI application fails to generate a log file; the `pactus.log` file is missing from the data folder.

### How To Reproduce

1. Download version v1.0.0 of the software.
2. Run the Windows-GUI application.
3. Observe that no log file is produced.

### What Happened

The root cause of this issue is the use of the `-H windowsgui` flag during the releasing process for the Windows binary, look  [here](https://github.com/pactus-project/pactus/blob/6f023154ef49861cb085b4d72ded4846ebbfc72e/.github/releasers/releaser_gui_windows.sh#L22C25-L22C39). 
This flag is documented [here](https://pkg.go.dev/cmd/link) and specifies that:

> On Windows, the `-H windowsgui` option produces a GUI binary instead of a console binary.

Using `-H windowsgui` detaches the console from the application. 
The logging mechanism works like this: Logs are written to `stderr`, and then the system reads the `stderr` output and writes it into a file. In the absence of a console, both `stdout` and `stderr` are nullified, resulting in no log file being generated.

### Solution

To resolve the issue, verify if `stderr` is null. If true, create a buffer writer to redirect log output from the logger to the file writer, ensuring log file creation.

### Environment

- **Operation System:** Windows
- **Pactus Version:** 1.0.0
- **Go Version:** 21.0

amirvalhalla:
## Description

Fix stderr logger in windows os.

## Related issue(s)

- Fixes #1077

### DIFF
--- a/util/logger/logger.go
+++ b/util/logger/logger.go
@@ -1,6 +1,7 @@
 package logger
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


b00f:
@amirvalhalla I have concern for memory leakage here.
To test it, change log levels of all modules to `trace`. You can also add an infinite loop to just write a log. Watch the memory of the system.

util/logger/logger.go
```diff
@@ -65,38 +66,48 @@ func getLoggersInst() *logger {
...
-	if globalInst == nil {
-		writers := []io.Writer{}
-		// console writer
+	if globalInst != nil {
+		return
+	}
+
+	writers := []io.Writer{}
+	// console writer
+	_, err := os.Stderr.Write([]byte{0})
```
b00f, line 75: // Check if stderr is null. If true, create a buffer writer to redirect log output from the logger to the file writer,
// ensuring log file creation: https://github.com/pactus-project/pactus/issues/1077

b00f: Use this comment instead

b00f (approved):

PR #1106, opened 2024-02-13
feat(grpc): update swagger API to version 1.1

b00f:
## Description

This PR updates the Swagger APIs to version 1.1. In this new version we follow this convention:

- Each API starts with `pactus`
- Each API should follow the same path as gRPC APIs
- Comments should be on top of the field to generate commented codes.


kehiy (approved):
Looks good. I just have one question. What's the purpose of the `pactus` prefix instead of `v1`, and why we removed the `v1` prefix? Isn't that helpful when we add the next versions over time?

b00f:
>  Isn't that helpful when we add the next versions over time?
I saw that swagger also has version, we bump it to v1.1 in this PR

kehiy:
what about SDKs and users that use these endpoints on code? they don't work with Swagger in code and they don't have access to the Swagger version.

b00f:
Even we add version to API, they still need to update their code. So this version is not really helpful for them

amirvalhalla (approved):

Issue #1108, opened 2024-02-13
Inacurate number of incound conections 


b00f:
## Description

Users have reported connection issues with their nodes. Upon investigation, we discovered that the count of inbound (and possibly outbound) connections is incorrect, significantly deviating from the actual count. For instance, in one node, the log shows:

```
{"_network":"{389}","inbound":1034,"message":"InterceptAccept rejected: many connections"}
```

This indicates a discrepancy where the node has only 389 connections but thinks it has 1034 inbound connections.

The problem arises from Libp2p occasionally failing to trigger the disconnect event, resulting in peers not being removed from PeerMgr. Look [here](https://github.com/pactus-project/pactus/blob/e63f1f8ac6b25ef78fbc967296800f43ba41caef/network/peermgr.go#L95). As a consequence, the count of inbound (or outbound?) connections becomes inaccurate, worsening over time.

One potential solution is to iterate through all connections in Libp2p->Host->Network to accurately determine the number of inbound and outbound connections. However, this approach raises concerns:

1. It is not efficient.
2. Testing different scenarios becomes challenging without network mocking (are there any solutions?).

Alternatively, we can iterate through the list of all peers in PeerMgr, checking their connectivity and removing those that are not connected. Although simpler, this method is still inefficient, albeit facilitating testing various scenarios.

PR #1109, opened 2024-02-13
fix(network): prevent adding connections with unknown direction to peer manager

kehiy:
## Description

We have an unknown direction in lp2ppeer and we need to prevent adding them in connections.

## Related issue(s)

- Related #1108

### DIFF
--- a/network/notifee.go
+++ b/network/notifee.go
@@ -91,6 +91,12 @@ func (s *NotifeeService) Connected(_ lp2pnetwork.Network, conn lp2pnetwork.Conn)
 	pid := conn.RemotePeer()
 	s.logger.Info("connected to peer", "pid", pid, "direction", conn.Stat().Direction, "addr", conn.RemoteMultiaddr())
 
+	if conn.Stat().Direction == lp2pnetwork.DirUnknown {
+		_ = conn.Close()
+
+		return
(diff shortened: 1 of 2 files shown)


network/notifee.go
```diff
@@ -91,6 +91,12 @@ func (s *NotifeeService) Connected(_ lp2pnetwork.Network, conn lp2pnetwork.Conn)
 	pid := conn.RemotePeer()
 	s.logger.Info("connected to peer", "pid", pid, "direction", conn.Stat().Direction, "addr", conn.RemoteMultiaddr())
 
+	if conn.Stat().Direction == lp2pnetwork.DirUnknown {
```
b00f, line 94: This is we need to investigate more.

kehiy: I think DirUnknown connections make problems for us. we can't determine whether they are inbound or outbound. In this way, we only accept known connections.

b00f: In go-libp2p, the `DirUnknown` flag has never been used. Therefore, we are adding code with the assumption that it will never be encountered.

kehiy: SO, if we are 100% sure it's not used. let's close it. I thought we have some unknown Dir connection.

PR #1118, opened 2024-02-14
feat(GUI): adding availability score in wallet

### LINKED ISSUE #912: Show availability score in Default wallet address GUI

In version 0.19 we add availability score for proposer validator, but client can't see validators currently score in GUI.

We can add new column `Score` for validators and show currently score of validator.

Screenshot 2024-01-07 154203

kehiy:
## Description

Screenshot from 2024-02-14 21-01-05

## Related issue(s)

- Fixes #912


b00f (commented):
Adding `GetAvailabilityScore` to wallet is not a good practice. 
We can pass `node` to `walletModel` to get validator info. Then we can even show if validators in unbonded or last sortition height as well

b00f (approved):

cmd/gtk/model_wallet.go
```diff
@@ -48,12 +54,23 @@ func (model *walletModel) rebuildModel() {
 			balanceStr := util.ChangeToString(balance)
 			stakeStr := util.ChangeToString(stake)
 
+			var score string
+
+			valAddr, err := crypto.AddressFromString(info.Address)
+			if err == nil {
+				val := model.node.State().ValidatorByAddress(valAddr)
```
b00f, line 61: val itself has `availibility_score`. right?

kehiy: ni, validator type does not contain an availability_score.

b00f (approved):

PR #1128, opened 2024-02-18
feat(wallet): adding all account address functions

kehiy:


wallet/vault/vault.go
```diff
@@ -232,6 +232,21 @@ func (v *Vault) AllValidatorAddresses() []AddressInfo {
 	return addrs
 }
 
+func (v *Vault) AllAccountAddresses() []AddressInfo {
+	addrs := make([]AddressInfo, 0, v.AddressCount()/2)
+	for _, addrInfo := range v.Addresses {
+		addrPath, _ := addresspath.NewPathFromString(addrInfo.Path)
+		if addrPath.AddressType() == H(crypto.AddressTypeBLSAccount) {
```
b00f, line 239:
```suggestion
		if addrPath.AddressType() != H(crypto.AddressTypeValidator) {
```

This is more accurate. In future we may support other curves.

b00f: Add a test for this function, in that tests the return addresses has no validator address

wallet/wallet.go
```diff
@@ -502,3 +506,24 @@ func (w *Wallet) AddTransaction(id tx.ID) error {
 func (w *Wallet) GetHistory(addr string) []HistoryInfo {
 	return w.store.History.getAddrHistory(addr)
 }
+
+func (w *Wallet) AllAmount(addr string, pt payload.Type) int64 {
```
b00f, line 510: Two notes:
1- The function name can change to "TotalAmount"
2- Let's add this in another PR. I think we may update the logic in a way to reduce the wasted changes.

b00f (changes requested):

wallet/wallet.go
kehiy: OK.

wallet/vault/vault_test.go
```diff
@@ -123,6 +123,23 @@ func TestSortAddressInfo(t *testing.T) {
...
 }
 
+func TestAllAccountAddresses(t *testing.T) {
+	td := setup(t)
+
+	assert.Equal(t, td.vault.AddressCount(), 6)
+
+	accountAddrs := td.vault.AllAccountAddresses()
+	for _, i := range accountAddrs {
+		info := td.vault.AddressInfo(i.Address)
```
b00f, line 133: can be simplified:
```
assert.NotEqual(t, H(crypto.AddressTypeValidator), addresspath.NewPathFromString(addrInfo.Path).AddressType()
```

Also, please rename `NewPathFromString` to `FromString`

b00f (approved):

PR #1137, opened 2024-02-23
feat: migrate wallet from json to sqlite

amirvalhalla:
## Description

Implementing sqlite for wallet.


amirvalhalla:
please don't merge the branch. just review it.

wallet2/db/db.go
```diff
@@ -0,0 +1,474 @@
...
+	}, nil
+}
+
+func (d *db) InsertIntoTransaction(t *Transaction) (*Transaction, error) {
+	insertQuery := fmt.Sprintf("INSERT INTO %s (tx_id, address, block_height, block_time,"+
+		" payload_type, data, description, amount, status, created_at) VALUES"+
+		" (?,?,?,?,?,?,?,?,?,?)", TransactionTable)
+
+	t.CreatedAt = time.Now().UTC()
+	r, err := d.ExecContext(context.Background(), insertQuery, t.TxID, t.Address, t.BlockHeight, t.BlockTime,
```
Ja7ad, line 170: Please use top level context for prevent database damage on execute query.

```go
 InsertIntoTransaction(ctx context.Context, t *Transaction) (*Transaction, error)
```

wallet2/db/db.go
```diff
@@ -0,0 +1,474 @@
...
+		PublicKey:  addr.PublicKey,
+		Label:      addr.Label,
+		Path:       addr.Path,
+		IsImported: addr.IsImported,
+		CreatedAt:  addr.CreatedAt,
+	}, nil
+}
+
+func (d *db) InsertIntoTransaction(t *Transaction) (*Transaction, error) {
+	insertQuery := fmt.Sprintf("INSERT INTO %s (tx_id, address, block_height, block_time,"+
```
Ja7ad, line 165: Please don't use fmt.Sprintf for avoiding sql injection: https://go.dev/doc/database/sql-injection

wallet2/db/db.go
Ja7ad, line 170: Use prepared query before execute

Ja7ad (changes requested):

wallet2/db/query.go
```diff
@@ -0,0 +1,17 @@
+package db
+
+import "fmt"
+
+type QueryOption string
+
+func WithTransactionStatus(status Status) QueryOption {
+	return QueryOption(fmt.Sprintf("WHERE status = %d", status))
```
Ja7ad, line 8: don't use fmt.Sprintf

Ja7ad (changes requested):

wallet2/db/query.go
amirvalhalla: what other option you suggest?

Ja7ad: please check: https://go.dev/doc/database/sql-injection

Ja7ad:
```go
func Query(query string, args ...any)
```

amirvalhalla: this status comes from `HardCodec Code`, so doesn't make any security issue.

Ja7ad: rows, err := db.Query("SELECT * FROM user WHERE id = ?", id)

wallet2/db/db.go
Ja7ad: https://pkg.go.dev/database/sql#DB.PrepareContext

wallet2/db/db.go
amirvalhalla: sql injection doesn't happen for this scenario because we're passing the table name from `hard coded section` and for other parameters we handled the sql injection.

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+type Pair struct {
+	Key       string
+	Value     string
+	CreatedAt time.Time
+}
+
+func NewDB(ctx context.Context, path string) (DB, error) {
+	dbInstance, err := sql.Open("sqlite", path)
```
Ja7ad, line 82: Please create here 'context.WithCancel(ctx)' and keep cancel function in DB struct.
Then create Close method and call cancel func

Ja7ad (changes requested):

wallet2/db/db.go
amirvalhalla: as we've talked before we stop it from wallet so the cancellation token will raise whenever wallet stop.

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	context.Context
+}
+
+type Address struct {
+	ID         int       `json:"id"`          // id of wallet
+	Address    string    `json:"address"`     // Address in the wallet
+	PublicKey  string    `json:"public_key"`  // Public key associated with the address
+	Label      string    `json:"label"`       // Label for the address
+	Path       string    `json:"path"`        // Path for the address
+	IsImported bool      `json:"is_imported"` // imported for purpose
```
b00f, line 57: @amirvalhalla why we need to save "IsImported"?
We can get it from Path. right?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	TxID        string    `json:"tx_id"`
+	Address     string    `json:"address"`
+	BlockHeight uint32    `json:"block_height"`
+	BlockTime   uint32    `json:"block_time"`
+	PayloadType string    `json:"payload_type"`
+	Data        string    `json:"data"`
+	Description string    `json:"description"`
+	Amount      int64     `json:"amount"`
+	Status      int       `json:"status"`
+	CreatedAt   time.Time `json:"created_at"`
```
b00f, line 72: I think Sqlite automatically adds this column.

wallet2/db/db.go
amirvalhalla: Yes, we can. I just added this field for the future to help us in the GUI to display imported paths faster.

wallet2/db/db.go
amirvalhalla: I will delete json tags

b00f: I mean "created_at" column.

wallet2/db/db.go
Ja7ad: We need to stop query when received interrupt signal.

For stop query you need send a cancellation signal via context cancel func.

wallet2/db/db.go
b00f: Is this the only reason?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
+package db
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strings"
+	"time"
+
+	//nolint:all
```
b00f, line 10: Remove this line and let fix the linting issue

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	_ "github.com/glebarez/go-sqlite"
+)
+
+const (
+	AddressTable     = "addresses"
+	TransactionTable = "transactions"
+	PairTable        = "pairs"
+)
+
+type DB interface {
```
b00f, line 20: Do we need to define interface here?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+
+	GetAddressByID(id int) (*Address, error)
+	GetAddressByAddress(address string) (*Address, error)
+	GetAddressByPath(p string) (*Address, error)
+
+	GetTransactionByID(id int) (*Transaction, error)
+	GetTransactionByTxID(id string) (*Transaction, error)
+
+	GetPairByKey(key string) (*Pair, error)
+	GetTotalRecords(tableName string, query string, args ...any) (int64, error)
```
b00f, line 37: The name is a bit confusing. 
If it returns the total record, why does it have a query parameter? 
So maybe we can rename it as "QueryTable."

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+
+	GetAllAddresses() ([]Address, error)
+	GetAllAddressesWithTotalRecords(pageIndex, pageSize int) ([]Address, int64, error)
```
b00f, line 40: A function can do many things is not recommended...

wallet2/db/db.go
b00f: Another confusion, why we have `GetTotalRecords` and `GetAllTransactions` 
They look doing the same thing

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	Description string    `json:"description"`
+	Amount      int64     `json:"amount"`
+	Status      int       `json:"status"`
+	CreatedAt   time.Time `json:"created_at"`
+}
+
+type Pair struct {
+	Key       string
+	Value     string
+	CreatedAt time.Time
```
b00f, line 78: Same here, I believe SQLite handle it.

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	if err := d.createTransactionTable(); err != nil {
+		return err
+	}
+
+	return d.createPairTable()
+}
+
+func (d *db) createAddressTable() error {
+	addressQuery := fmt.Sprintf("CREATE TABLE %s (id INTEGER PRIMARY KEY AUTOINCREMENT,"+
+		" address VARCHAR, public_key VARCHAR, label VARCHAR, path VARCHAR, is_imported BOOLEAN, created_at TIMESTAMP)",
```
b00f, line 107: Do we have fixed string size instead of VARCHAR in SQL?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+		AddressTable)
+
+	_, err := d.ExecContext(d.Context, addressQuery)
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
b00f, line 111: This code is not lovely! Any other way to check if a table exists?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	GetAllAddresses() ([]Address, error)
+	GetAllAddressesWithTotalRecords(pageIndex, pageSize int) ([]Address, int64, error)
+
+	GetAllTransactions(query string, args ...any) ([]Transaction, error)
+	GetAllTransactionsWithTotalRecords(pageIndex, pageSize int, query string, args ...any) ([]Transaction, int64, error)
+}
+
+type db struct {
+	*sql.DB
+	context.Context
```
b00f, line 48: To be same as other part of the code, let's give a name to it: `ctx`

wallet2/db/query.go
```diff
@@ -0,0 +1,19 @@
+package db
+
+const (
+	EmptyQuery = ""
+)
+
+type QueryOption string
+
+func WithTransactionStatus() string {
```
b00f, line 9: Maybe like
`func (qo *QueryOption) WithTransactionStatus() *QueryOption {...}`

b00f (changes requested):

wallet2/db/db.go
amirvalhalla: I don't think so, let me check it.

wallet2/db/db.go
amirvalhalla: what function do you mean?

b00f: `GetAllAddressesWithTotalRecords` It gets all address and calculates the total records...

wallet2/db/db.go
amirvalhalla: it doesn't automatically add `created_at` column

wallet2/db/db.go
amirvalhalla: it would be better to have, because we have some private methods which we don't want to expose them to higher layer of application plus for tests we don't have any problem because `newDB` method has input which we can pass `memory` then run all tests.

wallet2/db/db.go
amirvalhalla: `GetTotalRecords ` just returns the total count of records which they exist in db. we'll use it for pagination but `GetAllTransactions` returns all transaction records.

wallet2/db/db.go
amirvalhalla: I think for this case it's normal it doesn't make any issue. anyway we can remove total records from here and call directly `GetTotalRecords` method.

wallet2/db/db.go
amirvalhalla: the driver doesn't handle it

wallet2/db/db.go
amirvalhalla: we can use `VARCHAR(N)`, which `N` is size of string. should we refactor this section?

wallet2/db/db.go
amirvalhalla: I've searched for alternatives, but unfortunately, it seems there's no solution except for this one

wallet2/db/db.go
amirvalhalla: resolved.

wallet2/db/query.go
amirvalhalla: I think this way which we've used is better and I should delete `QueryOption` type. do you agree?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+type Address struct {
```
b00f, line 50: Why did you rename it to `Address`?
Don't you think `AddressInfo` make more sense?

wallet2/db/db.go
b00f: Anyway we can construct `AddressPath` once we reading from Database, then we can always check path for this purpose.

wallet2/db/db.go
b00f: `N` is size or max_size?

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	AddressTable     = "addresses"
+	TransactionTable = "transactions"
+	PairTable        = "pairs"
+)
+
+type DB interface {
+	CreateTables() error
+
+	InsertIntoAddress(addr *Address) (*Address, error)
+	InsertIntoTransaction(t *Transaction) (*Transaction, error)
```
b00f, line 23: please avoid single letter for var names.

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+	InsertIntoPair(key string, value string) (*Pair, error)
+
+	UpdateAddressLabel(addr *Address) (*Address, error)
+
+	GetAddressByID(id int) (*Address, error)
+	GetAddressByAddress(address string) (*Address, error)
+	GetAddressByPath(p string) (*Address, error)
```
b00f, line 30: same here, no single letter

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+
+	GetTransactionByID(id int) (*Transaction, error)
+	GetTransactionByTxID(id string) (*Transaction, error)
+
+	GetPairByKey(key string) (*Pair, error)
```
b00f, line 35: Can we isolated Pair from outside of this package?
Then we can have `func GetValue(key string) (string, error)`

wallet2/db/db.go
```diff
@@ -0,0 +1,549 @@
...
+type DB interface {
+	CreateTables() error
+
+	InsertIntoAddress(addr *Address) (*Address, error)
+	InsertIntoTransaction(t *Transaction) (*Transaction, error)
+	InsertIntoPair(key string, value string) (*Pair, error)
+
+	UpdateAddressLabel(addr *Address) (*Address, error)
+
+	GetAddressByID(id int) (*Address, error)
```
b00f, line 28: I don't see any benefit by exporting ID or Index. Address string should be unique always.

b00f (changes requested):

wallet2/db/db.go
amirvalhalla: N is max size. e.g. if we say `VARCHAR(50)` the maximum Len could be 50.

Ja7ad:
@b00f Currently we don't need to migrate to sqlite and json worked.

Close it?

b00f:
Yes, we can close it

PR #1139, opened 2024-03-02
chore: add bootstrap address for anpjwq

hain64:
## Description

Adding bootstrap address for anpjwq


b00f:
@hain64 Merged with thanks

PR #1154, opened 2024-03-09
fix: grpc wallet creation

ChandlerFerry:
## Description

Properly passes through the wallet Password rather than passing Language in as the Password.

https://github.com/pactus-project/pactus/blob/main/www/grpc/wallet.go#L65
`wallet.Create(walletPath, req.Mnemonic, req.Language, s.chainType)`
https://github.com/pactus-project/pactus/blob/main/wallet/wallet.go#L68
`func Create(walletPath, mnemonic, password string, chain genesis.ChainType)`

## Related issue(s)

- Fixes private discord discussion problem. https://discord.com/channels/@me/1215218453774864384/121580495175051269

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -413,7 +413,7 @@ func (w *Wallet) ImportPrivateKey(password string, prv *bls.PrivateKey) error {
 	return w.store.Vault.ImportPrivateKey(password, prv)
 }
 
-func (w *Wallet) PrivateKey(password, addr string) (crypto.PrivateKey, error) {
+func (w *Wallet) PrivateKey(password string, addr string) (crypto.PrivateKey, error) {
 	keys, err := w.store.Vault.PrivateKeys(password, []string{addr})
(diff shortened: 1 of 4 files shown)


wallet/wallet.go
```diff
@@ -98,10 +98,10 @@ func Create(walletPath, mnemonic, password string, chain genesis.ChainType) (*Wa
 	if err != nil {
 		return nil, err
 	}
-	err = vlt.UpdatePassword("", password)
-	if err != nil {
-		return nil, err
-	}
+	// err = vlt.UpdatePassword("", password)
```
b00f, line 101: This means you disabled password for all wallets. 
Any reason for that?

b00f (changes requested):

b00f:
@ChandlerFerry 
Please update the issue. The link directs to a private chat that is not accessible by the public. It's better if we mention the issue #1154 there.

PR #1169, opened 2024-03-13
chore: update bootstrap address for Stilucky

stilucky:
## Description

Updating bootstrap address for Stilucky


b00f:
@stilucky Merged with thanks

PR #1172, opened 2024-03-14
fix: showing fee in atmoic unit on HTTP transaction page

kehiy:
## Description

Example of the current version:
Screenshot from 2024-03-14 19-09-48


themantre:
This part is under refactor. Soon it will be done

PR #1174, opened 2024-03-16
refactor: introduce Amount data type for converting PAC units

b00f:
## Description

This PR refactors all amounts to use the `Amount` data type. It encapsulates the amount and can be used for converting between PAC units.


cmd/gtk/utils.go
```diff
@@ -156,7 +156,7 @@ func updateValidatorHint(lbl *gtk.Label, addr string, w *wallet.Wallet) {
 	if err != nil {
 		updateHintLabel(lbl, "")
 	} else {
-		hint := fmt.Sprintf("stake: %v", util.ChangeToString(stake))
+		hint := fmt.Sprintf("stake: %s", stake)
```
amirvalhalla, line 159: doesn't need to call `stake.String()`?

cmd/gtk/utils.go
```diff
@@ -171,7 +171,7 @@ func updateAccountHint(lbl *gtk.Label, addr string, w *wallet.Wallet) {
 	if err != nil {
 		updateHintLabel(lbl, "")
 	} else {
-		hint := fmt.Sprintf("balance: %v", util.ChangeToString(balance))
+		hint := fmt.Sprintf("balance: %s", balance)
```
amirvalhalla, line 174: same for here `balance.String()`?

cmd/wallet/tx.go
```diff
@@ -37,27 +37,30 @@ func buildTransferTxCmd(parentCmd *cobra.Command) {
...
+		trx, err := w.MakeTransferTx(from, to, amt, opts...)
 		cmd.FatalErrorCheck(err)
 
 		cmd.PrintLine()
 		cmd.PrintInfoMsgf("You are going to sign this \033[1mTransfer\033[0m transition:")
 		cmd.PrintInfoMsgf("From  : %s", from)
 		cmd.PrintInfoMsgf("To    : %s", to)
-		cmd.PrintInfoMsgf("Amount: %.9f", util.ChangeToCoin(amount))
-		cmd.PrintInfoMsgf("Fee   : %.9f", util.ChangeToCoin(trx.Fee()))
+		cmd.PrintInfoMsgf("Amount: %s", amt)
```
amirvalhalla, line 62: I think you should call `amt.String()`, right?

cmd/wallet/tx.go
```diff
@@ -79,27 +82,30 @@ func buildBondTxCmd(parentCmd *cobra.Command) {
...
+		trx, err := w.MakeBondTx(from, to, *pubKeyOpt, amt, opts...)
 		cmd.FatalErrorCheck(err)
 
 		cmd.PrintLine()
 		cmd.PrintInfoMsgf("You are going to sign this \033[1mBond\033[0m transition:")
 		cmd.PrintInfoMsgf("Account  : %s", from)
 		cmd.PrintInfoMsgf("Validator: %s", to)
-		cmd.PrintInfoMsgf("Stake    : %.9f", util.ChangeToCoin(amount))
-		cmd.PrintInfoMsgf("Fee      : %.9f", util.ChangeToCoin(trx.Fee()))
+		cmd.PrintInfoMsgf("Stake    : %s", amt)
```
amirvalhalla, line 107: same for here

types/validator/validator.go
```diff
@@ -106,16 +107,16 @@ func (val Validator) Power() int64 {
 		return 1
 	}
 
-	return val.data.Stake
+	return int64(val.data.Stake)
```
amirvalhalla, line 110: should use `amount.Amount(val.data.Stake)`

sandbox/mock.go
```diff
@@ -41,7 +41,7 @@ func MockingSandbox(ts *testsuite.TestSuite) *MockSandbox {
 		TestCommittedTrxs:    make(map[tx.ID]*tx.Tx),
 	}
 
-	treasuryAmt := amount.Amount(21000000 * 1e9)
+	treasuryAmt := amount.Amount(21_000_000 * 1e9)
```
kehiy, line 44: can we have our decimals/conversion factor as a global parameter?

amirvalhalla (changes requested):

amirvalhalla (approved):

sandbox/mock.go
b00f: This is only used in tests (note that it is not the total coins in network). Defining it globally for tests at this time is not a good idea unless we change the mocking structure for the sandbox. There is no good place for this global definition.
By the way, it is a good suggestion. We may work on this on another PR.

kehiy: No, not about coins amount. I'm talking about PAC coin decimals. I think we can define it globally.

b00f: We try to work on that in other PRs.

PR #1180, opened 2024-03-25
feat(GUI): memo field for transactions on GUI wallet

kehiy:
## Description

This PR will add a new field called memo in the wallet GUI. With this change, users can specify the memo of their transaction in GUI.


b00f:
@kehiy Please use `Amount` data type.

b00f:
Also better to rebase this branch.

kehiy:
@b00f amount data type is already used, my branch was too old and not synced. also, I have serious issued with git in these branches. please check my next PR.

PR #1190, opened 2024-04-02
feat(gRPC): add new API to get the total balance of wallet

b00f:
## Description

This PR adds a new API to gRPC endpoints to get the total balance of a wallet.


kehiy (approved):
LGTM. I think we need to add total balance of PACs in one address including fee of transferring.

then we can provide total option on wallets.

PR #1192, opened 2024-04-02
fix(wallet): using public node for transactions in first attempt

kehiy:
## Description

Right now Pactus only has 1 GUI wallet that can keep and transfer PAC. if we try to broadcast transactions to our node on the first attempt it will fail always and only those who use this software as a Node can use this wallet, with this change this GUI can be used only as a wallet.


b00f:
@kehiy 
GUI is attached to its state. This hack will cause many side effects and it is too dangerous.
If user's node is not synced they can still use `pactus-wallet` for sending transaction.

This PR can't be merged.

b00f:
I think if you test these changes, you will notice that the GUI will become slow because it retrieves wallet info from a remote server.

kehiy:
@b00f We should have solutions for that, since not everybody can work with cli wallet. 

for example, we are paying pactus coins to our team on dezh tech, and we can non-tech people there, how they can interact with pactus?

PR #1196, opened 2024-04-04
fix(GUI): make transaction hash selactable

b00f:
## Description

This PR updates MessageDilaog and enable MarkUp and set content selectable.
It also update Memo field position and label.


cmd/gtk/utils.go
```diff
@@ -19,9 +19,25 @@ import (
 	"github.com/pactus-project/pactus/wallet"
 )
 
+// https://stackoverflow.com/questions/3249053/copying-the-text-from-a-gtk-messagedialog
+func updateMessageDialog(dlg *gtk.MessageDialog) {
+	area, err := dlg.GetMessageArea()
```
kehiy, line 24: @b00f for address details we added CSS style and function that user can copy value of a field, can't we use it here too?

cmd/gtk/assets/ui/dialog_transaction_bond.ui
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.40.0 -->
+<!-- Generated with glade 3.38.2 -->
```
kehiy, line 2: small note: you can use glade 3.40.0 too.

kehiy (approved):

cmd/gtk/utils.go
b00f: Note sure if it can help there.

PR #1203, opened 2024-04-07
fix(GUI): updating total balance each 10 seconds

kehiy:
## Description

Our total balance won't be updated, with this change we run a goroutine that updates this balance every 10 seconds (per block).

### DIFF
--- a/cmd/gtk/widget_wallet.go
+++ b/cmd/gtk/widget_wallet.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	_ "embed"
+	"time"
 
 	"github.com/gotk3/gotk3/gdk"
 	"github.com/gotk3/gotk3/glib"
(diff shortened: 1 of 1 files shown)


cmd/gtk/widget_wallet.go
```diff
@@ -67,7 +68,13 @@ func buildWidgetWallet(model *walletModel) (*widgetWallet, error) {
 	} else {
 		labelEncrypted.SetText("No")
 	}
-	labelTotalBalance.SetText(model.wallet.TotalBalance().String())
+
+	go func() {
+		for {
+			labelTotalBalance.SetText(model.wallet.TotalBalance().String())
+			time.Sleep(10 * time.Second)
```
ambersun1234, line 75: I would prefer using `Timer`, please have a look at this [Go by Example](https://gobyexample.com/)

Ja7ad: @kehiy please use time.Ticker or time.Timer with for-select instead thread sleep.

kehiy: @ambersun1234 @Ja7ad 

The Time Ticker is used for having more control over our job when we are in a concurrent system, for example shutting down the job and ...

In our case, it's just a simple update on a balance and we won't need to do more complex stuff in this routine in the future as you know.

So, is there any good reason behind using tickers?

ambersun1234: It's all about efficiency, using time.sleep is like a spin lock, it'll waste cpu time.
timer on the other hand will not, the goroutine will wake up only when the timer expires

b00f: There is a potential for sudden panic in this code. Calling GTK APIs from a non-main thread could crash the software. This is explained [here](https://github.com/gotk3/gotk3/issues/686)

To address this issue, consider referring to PR #1204.

PR #1212, opened 2024-04-10
Fix grpc get block

b00f:
## Description

Block data and transaction data are omitted when the verbosity level is set to "information."


Ja7ad (approved):

www/grpc/blockchain.go
```diff
@@ -151,7 +155,11 @@ func (s *blockchainServer) GetBlock(_ context.Context,
 		trxs := make([]*pactus.TransactionInfo, 0, block.Transactions().Len())
 		for _, trx := range block.Transactions() {
 			if req.Verbosity == pactus.BlockVerbosity_BLOCK_INFO {
-				trxs = append(trxs, &pactus.TransactionInfo{Id: trx.ID().Bytes()})
+				data, _ := trx.Bytes()
+				trxs = append(trxs, &pactus.TransactionInfo{
+					Id:   trx.ID().Bytes(),
+					Data: data,
```
ambersun1234, line 161: shouldn't `BLOCK_INFO` only contain transaction id?

PR #1219, opened 2024-04-14
chore: bumping version to 1.2.0

ambersun1234:
Bumping version to 1.2.0


b00f:
We should update `releasing.md` as well

PR #1222, opened 2024-04-14
chore: bumping version to 1.2.0

ambersun1234:
Bumping version to 1.2.0


b00f:
In `releasing.md` we need to change the variables to:
```bash
PRV_VER="1.1.0"
CUR_VER="1.2.0"
NEXT_VER="1.3.0"
```

b00f:
This part should be simplified. I am thinking to change it

b00f (approved):

PR #1264, opened 2024-05-06
feat(gRPC): add only_connected parameter to getNetworkInfo API

kehiy:
## Description

This PR updates the Network API in gRPC to display only connected peers. Connected peers are those with a status of "Connected" or "Known".


www/grpc/network.go
```diff
@@ -48,12 +48,16 @@ func (s *networkServer) GetNodeInfo(_ context.Context,
...
 func (s *networkServer) GetNetworkInfo(_ context.Context,
-	_ *pactus.GetNetworkInfoRequest,
+	req *pactus.GetNetworkInfoRequest,
 ) (*pactus.GetNetworkInfoResponse, error) {
 	ps := s.sync.PeerSet()
 	peerInfos := make([]*pactus.PeerInfo, 0, ps.Len())
 
 	ps.IteratePeers(func(peer *peerset.Peer) bool {
+		if req.OnlyOnline && (!peer.IsConnected() || !peer.IsKnown()) {
+			return true
```
b00f, line 58: `return true` will break the iteration (stop it).
`return false` will simply continue it.

This code will break the loop at the first occurrence. Have you tested it?

www/grpc/network.go
```diff
@@ -48,12 +48,16 @@ func (s *networkServer) GetNodeInfo(_ context.Context,
...
 
 func (s *networkServer) GetNetworkInfo(_ context.Context,
-	_ *pactus.GetNetworkInfoRequest,
+	req *pactus.GetNetworkInfoRequest,
 ) (*pactus.GetNetworkInfoResponse, error) {
 	ps := s.sync.PeerSet()
 	peerInfos := make([]*pactus.PeerInfo, 0, ps.Len())
 
 	ps.IteratePeers(func(peer *peerset.Peer) bool {
+		if req.OnlyOnline && (!peer.IsConnected() || !peer.IsKnown()) {
```
b00f, line 57:
```suggestion
		if req.OnlyOnline && !peer.IsConnected() {
```

`IsConnected()` checks for both `Connected` and `Known` status

b00f (changes requested):

b00f:
@kehiy 
Please test it.
Can you add HTTP page for only online peers: like "/network/connected"

www/grpc/proto/network.proto
```diff
@@ -14,7 +14,10 @@ service Network {
 }
 
 // Request message for retrieving overall network information.
-message GetNetworkInfoRequest {}
+message GetNetworkInfoRequest {
+  // Only returns the peers with online status
+  bool only_online = 1;
```
b00f, line 19: `only_connected` makes more sense because other peers may be online but not connected to this peer.

b00f (changes requested):

www/grpc/network.go
kehiy: I tested it and it worked. but since I was running the node recently, maybe all of my connections disconnected. I'll change this condition.

kehiy:
@b00f HTTP is supported for this flag too. we can use it like:

`http://localhost:8081/network?onlyConnected=true`

kehiy:
@b00f can you have a review?

b00f (approved):

PR #1265, opened 2024-05-06
feat(daemon): add export feature create snapshot from blockchain data

Ja7ad:


b00f:
@Ja7ad We need PIP first and the export is better to be a Python script.
Since it is not for users to work with export command

PR #1270, opened 2024-05-09
fix(api): add swagger schemes

Ja7ad:
image


Ja7ad:
@b00f default schemes is on http

b00f (approved):

PR #1283, opened 2024-05-28
feat(network): ratelimit gossip for prevent spamming

Ja7ad:


network/ratelimit.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"sync"
+	"time"
+)
+
+type rateLimit struct {
+	referenceTime time.Time
+	threshold     uint8
+	counter       uint8
+	mu            sync.Mutex
```
b00f, line 12: We don't need mutex here.

network/ratelimit.go
```diff
@@ -0,0 +1,57 @@
...
+	r.referenceTime = time.Now()
+}
+
+func (r *rateLimit) increment() bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	// Check if the window has expired and reset if necessary
+	if time.Since(r.referenceTime) > r.window {
+		r.counter = 0
```
b00f, line 45: we can call `r.reset()`

network/ratelimit.go
```diff
@@ -0,0 +1,57 @@
...
+	r.counter = 0
+	r.referenceTime = time.Now()
+}
+
+func (r *rateLimit) increment() bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	// Check if the window has expired and reset if necessary
+	if time.Since(r.referenceTime) > r.window {
```
b00f, line 44:
```suggestion
	if r.diff() > r.window {
```

b00f (changes requested):

network/ratelimit.go
```diff
@@ -0,0 +1,57 @@
...
+		r.counter = 0
+		r.referenceTime = time.Now()
+	}
+
+	r.counter++
+
+	// Check if the threshold is exceeded
+	if r.counter > r.threshold {
```
b00f, line 52: Simply we can replace these lines with:
	`return r.counter <= r.threshold`

b00f (approved):

PR #1298, opened 2024-05-31
feat(network): add firewall black list ips to banning when connecting to peer

Ja7ad:
# Description

This PR defines a blacklist that contains IP addresses that should be blocked. When a node attempts to connect, its IP address is checked against the blacklist. If it matches an entry, the connection is rejected.


sync/firewall/config.go
```diff
@@ -1,16 +1,42 @@
...
+//go:embed black_list.json
+var _defaultBlackListIPs []byte
+
 type Config struct {
-	Enabled bool `toml:"enable"`
+	Enabled      bool     `toml:"enable"`
+	BlackListIPs []string `toml:"blacklist_ips"`
+}
+
+type defaultBlackListIPs struct {
```
b00f, line 16: Why we have defaulted here?

sync/firewall/black_list.json
```diff
@@ -0,0 +1,5 @@
+{
+  "ips": [
+    "115.193.157.138"
```
b00f, line 3: Looks redundant to me

sync/firewall/firewall_test.go
```diff
@@ -78,6 +84,16 @@ func TestInvalidBundlesCounter(t *testing.T) {
 	assert.Equal(t, p.InvalidBundles, 3)
 }
 
+func TestBlackListIPs(t *testing.T) {
+	td := setup(t)
+
+	blackListIP := "10.10.10.10"
```
b00f, line 90: Remote address is in format of libp2p address. Update test

b00f (changes requested):

sync/firewall/config.go
Ja7ad: This Default append to configuration list, some time we find faulty similar now we want banned on patch version.

Same now.

b00f (approved):

PR #1308, opened 2024-06-02
feat(firewall): define rate limit for network topics

b00f:
## Description

This PR defines rate limits for network topics. 
The rate limit specifies the number of messages per second that are allowed. 
If the threshold is exceeded, the message won't be broadcasted.
It also define new topics for block and transactions.


network/gossip.go
```diff
@@ -12,21 +13,23 @@ import (
...
-	generalTopicName string
-	rateLimit        *rateLimit
+	ctx                    context.Context
+	wg                     sync.WaitGroup
+	config                 *Config
+	host                   lp2phost.Host
+	pubsub                 *lp2pps.PubSub
+	topics                 []*lp2pps.Topic
+	subs                   []*lp2pps.Subscription
+	generalTopicDeprecated *lp2pps.Topic
```
Ja7ad, line 23: Instead rename topic to deprecated you can use standard comment for deprecated, for example:

```go
generalTopic *lp2pps.Topic // Deprecated: generalTopic ...
```

network/gossip.go
```diff
@@ -58,15 +61,130 @@ func newGossipService(ctx context.Context, host lp2phost.Host, eventCh chan Even
...
+		}
+
+		return g.BroadcastMessage(msg, g.consensusTopic)
+
+	default:
+		return InvalidTopicError{TopicID: topicID}
+	}
+}
+
+func (g *gossipService) JoinGeneralTopic(sp ShouldPropagate) error {
```
Ja7ad, line 110: Add comment Deprecated for this method:

```go
// Deprecated: JoinGeneralTopic
func (g *gossipService) JoinGeneralTopic(sp ShouldPropagate) error {}
```

network/gossip.go
```diff
@@ -58,15 +61,130 @@ func newGossipService(ctx context.Context, host lp2phost.Host, eventCh chan Even
...
+	topic, err := g.JoinTopic(g.consensusTopicName(), sp)
+	if err != nil {
+		return err
+	}
+	g.consensusTopic = topic
+
+	return nil
+}
+
+func (g *gossipService) generalTopicNameDeprecated() string {
```
Ja7ad, line 170: Please add comment deprecated instead rename method:

```go
// Deprecated: generalTopicName
func (g *gossipService) generalTopicName() string {}
```

network/interface.go
```diff
@@ -9,15 +9,23 @@ import (
 type TopicID int
 
 const (
-	TopicIDGeneral   TopicID = 1
-	TopicIDConsensus TopicID = 2
+	TopicIDGeneralDeprecated TopicID = -1
```
Ja7ad, line 12: Please add comment deprecated instead rename enum:

```go
TopicIDGeneral   TopicID = -1 // Deprecated: TopicIDGeneral
```

Ja7ad (changes requested):

network/gossip.go
b00f: Done.

network/gossip.go
b00f: Done.

network/interface.go
b00f: Done.

Ja7ad (approved):

Ja7ad:
@b00f Thank you

PR #1309, opened 2024-06-02
fix(firewall): block peer base on blacklisted CIDR

Ja7ad:
# Description

This PR allows nodes to block peers based on a given CIDR range, instead of individual IP addresses. This enables each node to block a range of IPs more effectively.


config/example_config.toml
```diff
@@ -93,7 +93,7 @@
 
   # `sync.firewall` contains configuration options for the sync firewall.
   [sync.firewall]
-    # `blacklist_addresses` contains the list of addresses that should be blacklisted.
+    # `blacklist_addresses` contains the list of ip cidrs that should be blacklisted.
```
b00f, line 96:
```suggestion
    # `blacklist_addresses` contains the list of IPs and CIDRs that should be blacklisted.
```

sync/firewall/firewall.go
```diff
@@ -26,13 +27,25 @@ type Firewall struct {
...
 func NewFirewall(conf *Config, net network.Network, peerSet *peerset.PeerSet, st state.Facade,
 	log *logger.SubLogger,
-) *Firewall {
-	conf.LoadDefaultBlackListAddresses()
+) (*Firewall, error) {
+	if err := conf.LoadDefaultBlackListAddresses(); err != nil {
+		return nil, err
+	}
+
+	if err := conf.BasicCheck(); err != nil {
```
b00f, line 41: We have this check once we run the node, look [here](https://github.com/pactus-project/pactus/blob/9c5d7bd4c379207ee6378149dcedbb5adeb0e90b/cmd/cmd.go#L385)
Please add this code inside here: https://github.com/pactus-project/pactus/blob/e947cb662e737233d2c91664f0558fb75d4454b4/sync/config.go#L43

sync/firewall/firewall.go
```diff
@@ -66,16 +80,13 @@ func (f *Firewall) OpenGossipBundle(data []byte, from peer.ID) *bundle.Bundle {
 }
 
 func (f *Firewall) IsBlackListAddress(remoteAddr string) bool {
-	p2pRemoteAddr, err := addr.Parse(remoteAddr)
+	ip, err := f.parseP2PAddr(remoteAddr)
 	if err != nil {
 		f.logger.Debug("firewall: unable to parse remote address", "err", err)
```
b00f, line 85:
```suggestion
		f.logger.Warn("firewall: unable to parse remote address", "err", err, "addr", remoteAddr)
```

sync/firewall/firewall.go
```diff
@@ -179,3 +190,27 @@ func (f *Firewall) AllowTransactionRequest() bool {
 func (f *Firewall) AllowConsensusRequest() bool {
 	return f.consensusRateLimit.AllowRequest()
 }
+
+func (f *Firewall) parseP2PAddr(address string) (string, error) {
```
b00f, line 194: Better to give a semantic name, like: `getIPfromMultiaddress`

util/ipblocker/ipblocker.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net"
+	"sync"
+)
+
+type IPBlocker struct {
+	cidrs []*net.IPNet
+}
+
+func New(blackListCidr []string) (*IPBlocker, error) {
```
b00f, line 12: To be in harmony with the rest of code, lets rename it to "NewIPBlocker"

util/ipblocker/ipblocker.go
```diff
@@ -0,0 +1,56 @@
...
+	if parsedIP == nil {
+		return false
+	}
+
+	var wg sync.WaitGroup
+	result := make(chan bool, 1)
+
+	for _, cidr := range i.cidrs {
+		wg.Add(1)
+		go func(cidr *net.IPNet) {
```
b00f, line 39: Why routine here? Is it blocking?

b00f (changes requested):

util/ipblocker/ipblocker.go
Ja7ad: It's standard name base on go effective.

```go
ipblocker.New()
```

Package name make explicit it, not required another things.

Because package encapsulated.

util/ipblocker/ipblocker.go
Ja7ad: Base on increased CIDRs and IPs we need use goroutine for fast checking.

b00f (approved):

util/ipblocker/ipblocker.go
b00f: @Ja7ad Please update other packages.

b00f (approved):

Issue #1322, opened 2024-06-06
Panic when get total account balance


Ja7ad:
## Description

When get total balance panic happen on nil conn.

### What Happened

![image](https://github.com/pactus-project/pactus/assets/56496801/e4ba7889-603c-4dc7-9d5a-91c20d0f0cb1)

b00f:
## How to Reproduce:

1. Build the latest code and run the Pactus blockchain locally. Localnet works as well. Read more [here](https://github.com/pactus-project/pactus/blob/main/docs/install.md).
2. In a separate terminal, run the following commands:
    ```
    ./build/pactus-shell wallet load-wallet --wallet-name "default_wallet"
    ./build/pactus-shell wallet get-total-balance --wallet-name "default_wallet"
    ```
3. The Pactus node crashes.

## How to Fix

1. First, ensure the node does not crash on gRPC API failure. Try to catch the panic and log it.
2. Secondly, fix the above crash.

b00f:
@akbariandev any progress on this issue?

PR #1330, opened 2024-06-09
fix(grpc): improve grpc server and client

Ja7ad:
## Description

- add GRPC recovery middleware to prevent crash server on grpc
- add GRPC retry middleware to reconnection client


b00f:
@Ja7ad Do you think this fix #1322 is partial? (Once we have a panic, the node won't terminate.)

Ja7ad:
> @Ja7ad Do you think this fix panic issue in #1322? (Once we have a panic, the node won't terminate.)

This middleware fix panic on all grpc methods, for issue #1322 we need to fix nil client.

b00f (approved):

PR #1335, opened 2024-06-10
test: add more logs for network-test

b00f:
## Description

This PR adds more logs for the NetworkTest to trace the test failure.

### DIFF
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -210,7 +210,7 @@ func TestNetwork(t *testing.T) {
 	time.Sleep(2 * time.Second)
 
 	t.Run("Supported Protocols", func(t *testing.T) {
-		fmt.Println(t.Name())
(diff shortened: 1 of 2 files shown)


network/network_test.go
```diff
@@ -230,6 +232,8 @@ func TestNetwork(t *testing.T) {
 	})
 
 	t.Run("all nodes have at least one connection to the bootstrap node B", func(t *testing.T) {
+		fmt.Println(t.Name())
```
Ja7ad, line 235: You can use `t.Log(t.Name())` instead fmt.Println.

Ja7ad (changes requested):

network/network_test.go
b00f: Done with thanks

Ja7ad (approved):

PR #1365, opened 2024-06-25
fix(network): use goroutines for sending streams

b00f:
## Description

We had a known issue that happened occasionally. Sometimes 
the nodes would lag for a minute and then start syncing. This PR fixes this issue. 
The main reason was that once `SendTo` blocked (for any reason), the receiving loop also got stuck. 
Here, we send messages asynchronously to ensure the receiving loop doesn't get blocked.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -368,14 +368,26 @@ func (n *network) Protect(pid lp2pcore.PeerID, tag string) {
 	n.host.ConnManager().Protect(pid, tag)
 }
 
-func (n *network) SendTo(msg []byte, pid lp2pcore.PeerID) error {
-	n.logger.Trace("Sending new message", "to", pid)
-
-	return n.stream.SendRequest(msg, pid)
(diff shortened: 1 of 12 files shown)


network/network.go
```diff
@@ -368,14 +368,26 @@ func (n *network) Protect(pid lp2pcore.PeerID, tag string) {
...
 
-func (n *network) Broadcast(msg []byte, topicID TopicID) error {
-	return n.gossip.Broadcast(msg, topicID)
+// Broadcast sends a message to all peers subscribed to a specific topic asynchronously.
+// It uses a goroutine to ensure that if broadcasting is blocked, receiving messages won't be blocked.
+func (n *network) Broadcast(msg []byte, topicID TopicID) {
+	go func() {
+		err := n.gossip.Broadcast(msg, topicID)
+		if err != nil {
+			n.logger.Error("error on broadcasting msg", "error", err)
```
themantre, line 388: Make it Warn

themantre (approved):

PR #1395, opened 2024-07-02
feat(wallet): query incremental timeout

kehiy:
## Description

This PR creates a time-out for querying data from gRPC on the wallet. we set a timeout for 5 seconds where the default is 30 seconds. we can make sure a slow RPC won't slow down the proccess.

### DIFF
--- a/wallet/client.go
+++ b/wallet/client.go
@@ -45,7 +45,6 @@ func (c *grpcClient) connect() error {
 		return nil
 	}
 
-	s := 1
 	for _, server := range c.servers {
 		conn, err := grpc.NewClient(server,
 			grpc.WithTransportCredentials(insecure.NewCredentials()),
@@ -54,26 +53,19 @@ func (c *grpcClient) connect() error {
 			continue
 		}
 
-		ctx, cancel := context.WithTimeout(c.ctx, time.Second*time.Duration(s))
-
 		blockchainClient := pactus.NewBlockchainClient(conn)
 		transactionClient := pactus.NewTransactionClient(conn)
 
 		// Check if client is responding
-		_, err = blockchainClient.GetBlockchainInfo(ctx,
+		_, err = blockchainClient.GetBlockchainInfo(c.ctx,
 			&pactus.GetBlockchainInfoRequest{})
 		if err != nil {
-			cancel()
-			s++
-
 			continue
 		}
 
 		c.conn = conn
 		c.blockchainClient = blockchainClient
 		c.transactionClient = transactionClient
-
-		cancel()
 
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


wallet/client.go
```diff
@@ -118,11 +129,14 @@ func (c *grpcClient) sendTx(trx *tx.Tx) (tx.ID, error) {
...
 
+	ctx, cancel := context.WithTimeout(c.ctx, time.Second*5)
+	defer cancel()
+
 	data, err := trx.Bytes()
 	if err != nil {
 		return hash.UndefHash, err
 	}
-	res, err := c.transactionClient.BroadcastTransaction(c.ctx,
+	res, err := c.transactionClient.BroadcastTransaction(ctx,
```
Ja7ad, line 139: Important things is gRPC method `BroadcastTransaction` should pass context to lower layers to stop process.

When you call gRPC make a go routine in background to processing gRPC method then your context send signal cancel base on timeout, process don't canceled in background.

Ja7ad (changes requested):
for create context not required to call `context.WithoutCancel`.

```go
ctx := context.WithoutCancel(context.Background())
```

kehiy:
> for create context not required to call `context.WithoutCancel`.
> 
> ```go
> ctx := context.WithoutCancel(context.Background())
> ```

I'm calling WithTimeout, not WithCancel.

wallet/client.go
kehiy: Which layer I should pass it to?

kehiy: What should we do to cancel it on background? also, why it won't cancel in background when the timeout reaches?

Ja7ad:
> > for create context not required to call `context.WithoutCancel`.
> > ```go
> > ctx := context.WithoutCancel(context.Background())
> > ```
> 
> I'm calling WithTimeout, not WithCancel.

line 29

kehiy:
> > > for create context not required to call `context.WithoutCancel`.
> > > ```go
> > > ctx := context.WithoutCancel(context.Background())
> > > ```
> > 
> > 
> > I'm calling WithTimeout, not WithCancel.
> 
> line 29

this line is not included in this PR, let me update it.

kehiy:
@Ja7ad fixed.

wallet/client.go
```diff
@@ -76,7 +78,10 @@ func (c *grpcClient) getBlockchainInfo() (*pactus.GetBlockchainInfoResponse, err
 		return nil, err
 	}
 
-	info, err := c.blockchainClient.GetBlockchainInfo(c.ctx,
+	ctx, cancel := context.WithTimeout(c.ctx, time.Second*5)
```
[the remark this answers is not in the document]
kehiy: It seems this way of handling the issue it not correct based on @Ja7ad review. once we find the correct way I'll cleanly handle the code. this PR needs change.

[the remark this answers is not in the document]
kehiy: @b00f @Ja7ad 
can you recheck this?
Is there any close method to call cancel there? or should I run a routine to call the cancel based on a condition?

Ja7ad:
@kehiy This method is not correct to solve the problem and I will implement another method to solve the problem.

PR #1396, opened 2024-07-03
feat(wallet): add timeout client connection

Ja7ad:
## Description

This PR introduces a client timeout for connections when dialing to the server. The primary goal is to prevent delays during query retrieval by setting a maximum duration for establishing a connection. This ensures that if a server is unresponsive or slow, the client will not hang indefinitely, thereby improving the overall responsiveness and reliability of the application.

### DIFF
--- a/cmd/wallet/main.go
+++ b/cmd/wallet/main.go
@@ -9,10 +9,10 @@ import (
 )
 
 var (
-	pathOpt       *string
-	offlineOpt    *bool
-	serverAddrOpt *string
-	timeoutOpt    *int
+	pathOpt        *string
+	offlineOpt     *bool
+	serverAddrsOpt *[]string
+	timeoutOpt     *int
 )
 
 func addPasswordOption(c *cobra.Command) *string {
@@ -21,16 +21,20 @@ func addPasswordOption(c *cobra.Command) *string {
 }
 
 func openWallet() (*wallet.Wallet, error) {
-	wlt, err := wallet.Open(*pathOpt, *offlineOpt)
+	opts := make([]wallet.Option, 0)
+
+	if *serverAddrsOpt != nil {
+		opts = append(opts, wallet.WithCustomServers(*serverAddrsOpt))
+	}
+
+	if *timeoutOpt > 0 {
+		opts = append(opts, wallet.WithTimeout(time.Duration(*timeoutOpt)*time.Second))
+	}
+
+	wlt, err := wallet.Open(*pathOpt, *offlineOpt, opts...)
 	if err != nil {
 		return nil, err
 	}
-
-	if *serverAddrOpt != "" {
-		wlt.SetServerAddr(*serverAddrOpt)
-	}
-
-	wlt.SetClientTimeout(time.Duration(*timeoutOpt) * time.Second)
 
 	return wlt, err
 }
@@ -48,12 +52,9 @@ func main() {
 	pathOpt = rootCmd.PersistentFlags().String("path",
 		cmd.PactusDefaultWalletPath(cmd.PactusDefaultHomeDir()), "the path to the wallet file")
 	offlineOpt = rootCmd.PersistentFlags().Bool("offline", false, "offline mode")
-	serverAddrOpt = rootCmd.PersistentFlags().String("server", "", "server gRPC address")
+	serverAddrsOpt = rootCmd.PersistentFlags().StringSlice("servers", []string{}, "servers gRPC address")
 	timeoutOpt = rootCmd.PersistentFlags().Int("timeout", 1,
-		"specifies the timeout duration for the client connection in seconds. "+
-			"this timeout determines how long the client will attempt to establish a connection with the server "+
-			"before failing with an error. Adjust this value based on network conditions and server response times "+
-			"to ensure a balance between responsiveness and reliability.")
+		"specifies the timeout duration for the client connection in seconds")
 
 	buildCreateCmd(rootCmd)
 	buildRecoverCmd(rootCmd)
--- a/cmd/wallet/main.go
+++ b/cmd/wallet/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"time"
+
 	"github.com/pactus-project/pactus/cmd"
 	"github.com/pactus-project/pactus/wallet"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


kehiy (commented):
Is this work in GUI wallet as well?

kehiy:
@Ja7ad I've tested it manually with 1 second as timeout but it took 24 seconds to respond to me.

My command:
```sh
./pactus-wallet address balance pc1zvchrej8nwfgvsduw47r4qmss5u25kvls2fx8e6 --timeout 1 
```

wallet/client.go
```diff
@@ -39,16 +43,30 @@ func (c *grpcClient) SetServerAddrs(servers []string) {
...
+		if len(c.servers) > 1 {
+			opts = append(opts, grpc.WithContextDialer(func(_ context.Context, s string) (net.Conn, error) {
+				return net.DialTimeout("tcp", s, c.timeout)
+			}))
+		}
+
+		conn, err := grpc.NewClient(server, opts...)
 		if err != nil {
-			continue
+			return err
```
kehiy, line 69: Here we must continue, we have a list of servers...
We can't stop the whole process if one of the servers is offline.

kehiy (changes requested):

Ja7ad:
> @Ja7ad I've tested it manually with 1 second as timeout but it took 24 seconds to respond to me.
> 
> My command:
> 
> ```shell
> ./pactus-wallet address balance pc1zvchrej8nwfgvsduw47r4qmss5u25kvls2fx8e6 --timeout 1 
> ```
> 
> Shouldn't it return a timeout error?

This timeout duration for dialing client, gRPC client is immediately and connecting the server happens in background.

When call `GetBlockchainInfo` client started to dialing and call RPC method.

Ja7ad:
> @Ja7ad so, with only 7 server, and 1 second timeout after 7 second I have to get data or timeout or can't connect to servers.
> 
> Here is the servers list:
> 
> ```json
> "mainnet": [
>         "localhost:50051",
>         "bootstrap1.pactus.org:50051",
>         "bootstrap2.pactus.org:50051",
>         "bootstrap3.pactus.org:50051",
>         "bootstrap4.pactus.org:50051",
>         "65.109.234.125:50051",
>         "157.90.111.140:50051"
> ]
> ```
> 
> But it took 24 seconds in my case, am I wrong?

This timeout is when dialing not query.

kehiy:
@Ja7ad our goal is to reduce the latency or just get a timeout. what do you think if we apply the timeout to queries as well? as we defined it on Trello task?

The rest looks good to me, we can merge it.

Ja7ad:
> @Ja7ad our goal is to reduce the latency or just get a timeout. what do you think if we apply the timeout to queries as well? as we defined it on Trello task?
> 
> The rest looks good to me, we can merge it.

Our delay is on when dialing in the best server.

kehiy:
@Ja7ad but based on my manual test, the process is still slow but there is no timeout error. which is not the same as our goal.

cmd/wallet/main.go
```diff
@@ -44,6 +49,11 @@ func main() {
 		cmd.PactusDefaultWalletPath(cmd.PactusDefaultHomeDir()), "the path to the wallet file")
 	offlineOpt = rootCmd.PersistentFlags().Bool("offline", false, "offline mode")
 	serverAddrOpt = rootCmd.PersistentFlags().String("server", "", "server gRPC address")
+	timeoutOpt = rootCmd.PersistentFlags().Int("timeout", 1,
+		"specifies the timeout duration for the client connection in seconds. "+
+			"this timeout determines how long the client will attempt to establish a connection with the server "+
```
b00f, line 54: It is too long. How the `-h` looks like after this change?
Do we have "Long Help" option in cobra?

wallet/client.go
```diff
@@ -20,15 +22,17 @@ type grpcClient struct {
...
 	transactionClient pactus.TransactionClient
 }
 
 func newGrpcClient() *grpcClient {
-	ctx := context.WithoutCancel(context.Background())
+	ctx := context.Background()
 
 	return &grpcClient{
 		ctx:               ctx,
+		timeout:           time.Second * 1,
```
b00f, line 35: Can't we get timeout when we instantiate the client?
```go 
func newGrpcClient(timeout int) {...}
````

File: wallet/client.go [hunk 4]
```diff
@@ -58,6 +76,10 @@ func (c *grpcClient) connect() error {
 		_, err = blockchainClient.GetBlockchainInfo(c.ctx,
 			&pactus.GetBlockchainInfoRequest{})
 		if err != nil {
+			if err := conn.Close(); err != nil {
```
b00f, line 79: I suggest to ignore error here:
```
_ = conn.Close()
```

wallet/wallet.go
```diff
@@ -156,6 +156,10 @@ func (w *Wallet) SetServerAddr(addr string) {
 	w.grpcClient.SetServerAddrs([]string{addr})
 }
 
+func (w *Wallet) SetClientTimeout(timeout time.Duration) {
```
b00f, line 159: Set timoeout when creating wallet. Then we don't need this setter.

wallet/client.go
```diff
@@ -39,14 +43,28 @@ func (c *grpcClient) SetServerAddrs(servers []string) {
...
 		return nil
 	}
 
 	for _, server := range c.servers {
-		conn, err := grpc.NewClient(server,
-			grpc.WithTransportCredentials(insecure.NewCredentials()))
+		opts := make([]grpc.DialOption, 0)
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+		if len(c.servers) > 1 {
```
b00f, line 61: There is one concern here:
1- User specify a server and timeout at the same time. This condition will not apply the timeout.
2- If `WalletManager`  creates wallet, it try to connect to `localhost`. In this case we may give large number for timeout.

b00f (changes requested):

cmd/wallet/main.go
b00f: I checked the Cobra. The 3rd argument in Options is `Usage`. So if we need to add Long Help we should add it into the command. It supports Long Help .

Ja7ad: > I checked the Cobra. The 3rd argument in Options is `Usage`. So if we need to add Long Help we should add it into the command. It supports Long Help .

This is switch not flag, just have usage.

Ja7ad:
@b00f done

b00f (approved):

Issue #1392, opened 2024-07-02
Pruning Store


Ja7ad:
A Full Node can convert to a Pruned Node by pruning the store and removing old blocks. Pruning the store is a time and resource consuming process and should not be performed while the node is running. We need to add a new command named prune to pactus-daemon to prune an offline node. A public method Prune should be defined in the store. This method iterates over all blocks from the pruning height to the genesis block and prunes them. The pruning height is LastBlockHeight - RetentionBlocks.

https://pips.pactus.org/PIPs/pip-27#pruning-store

PR #1399, opened 2024-07-08
feat(store): prune block function

### LINKED ISSUE #1391: PruneBlock function
The PruneBlock function removes a block and all transactions inside the block from the store. It is a private function and can’t be accessed from outside the store. It accepts a Batch pointer and a block height, and returns a boolean indicating whether a block at the given height exists, and an error if there is any. The PruneBlock function first tries to retrieve the block at the given height and decode it. Once it is decoded, it iterates over all transactions and updates the batch by deleting the associated keys from the store. Using a batch ensures atomicity [1](https://pips.pactus.org/PIPs/pip-27#fn:1) and improves performance by executing multiple delete operations as a single unit.

https://pips.pactus.org/PIPs/pip-27#pruneblock-function

kehiy:
## Description

This PR implements the prune block function defined on PIP-27.

## Related issue

- Fixes #1391

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -379,7 +379,7 @@ func (s *store) IsBanned(addr crypto.Address) bool {
 	return s.config.BannedAddrs[addr]
 }
 
-func (s *store) pruneBlock(batch *leveldb.Batch, blockHeight uint32) (bool, error) { //nolint
+func (s *store) pruneBlock(blockHeight uint32) (bool, error) { //nolint
 	s.lk.Lock()
 	defer s.lk.Unlock()
 
@@ -398,7 +398,7 @@ func (s *store) pruneBlock(batch *leveldb.Batch, blockHeight uint32) (bool, erro
 	}
 
 	for _, t := range blk.Transactions() {
-		batch.Delete(t.ID().Bytes())
+		s.batch.Delete(t.ID().Bytes())
 	}
 
 	return true, nil
--- a/store/store.go
+++ b/store/store.go
@@ -378,3 +378,28 @@ func (s *store) WriteBatch() error {
 func (s *store) IsBanned(addr crypto.Address) bool {
 	return s.config.BannedAddrs[addr]
 }
+
+func (s *store) pruneBlock(batch *leveldb.Batch, blockHeight uint32) (bool, error) { //nolint
+	s.lk.Lock()
+	defer s.lk.Unlock()
+
+	blkData, err := s.blockStore.block(blockHeight)
+	if err != nil {
+		return false, err
+	}
+
+	blk, err := block.FromBytes(blkData)
+	if err != nil {
+		return false, err
+	}
+
+	if err := s.blockStore.db.Delete(blk.Hash().Bytes(), nil); err != nil {
+		return false, err
+	}
+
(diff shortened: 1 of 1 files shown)


store/store.go
```diff
@@ -378,3 +378,28 @@ func (s *store) WriteBatch() error {
 func (s *store) IsBanned(addr crypto.Address) bool {
 	return s.config.BannedAddrs[addr]
 }
+
+func (s *store) pruneBlock(batch *leveldb.Batch, blockHeight uint32) (bool, error) { //nolint
```
b00f, line 382: why `//nolint`?

Ja7ad, line 382: You can get batch from `s *store` receiver, just get block height as parameter.

Ja7ad (changes requested):

store/store.go
```diff
@@ -378,3 +378,28 @@ func (s *store) WriteBatch() error {
...
+	}
+
+	if err := s.blockStore.db.Delete(blk.Hash().Bytes(), nil); err != nil {
+		return false, err
+	}
+
+	for _, t := range blk.Transactions() {
+		batch.Delete(t.ID().Bytes())
+	}
+
```
b00f, line 403: We should remove `blockHashKey`

b00f (changes requested):

store/store.go
Ja7ad, line 403: Please add `s.batch.Reset()` after prune block with transactions.

Ja7ad (changes requested):

store/store.go
b00f: Why batch should `Reset`?

Ja7ad: > Why batch should `Reset`?

I don't have good reason for this but check leveldb tests, after batch operation, reset used.

https://github.com/syndtr/goleveldb/blob/126854af5e6d8295ef8e8bee3040dd8380ae72e8/leveldb/batch_test.go#L87

store/store.go
kehiy: It's a private function and currently its not used. once it used by other functions we can remove it.

kehiy: It's defined at issue details. wierd....
let me update it.

kehiy: fixed

store/store.go
kehiy: In TXs? or you mean I need to remove the block using its hash key?

kehiy: @Ja7ad Maybe they are using this to prevent conflicts in test(?).
We had other batch operations on the store without calling reset.

b00f: We should have a reason for that.

b00f: Check `blockHashKey` in this file. We can get block hash, by block height.
When we remove the block, this record can be deleted.

b00f (changes requested):

Ja7ad:
@b00f For `prune` command in task #1392 we need to get public method to prune 1 block with transactions in loop.

b00f (approved):

b00f:
> @b00f For `prune` command in task #1392 we need to get public method to prune 1 block with transactions while pruning.

the `pruneBlock` is private

PR #1400, opened 2024-07-08
feat(core): pruning client by prune command

### LINKED ISSUE #1392: Pruning Store

Ja7ad:
## Description

The prune command optimizes blockchain storage by removing outdated blocks and transactions, freeing up disk space and enhancing client performance.

## Related issue(s)

- Fixes #1392

### DIFF
--- a/cmd/daemon/prune.go
+++ b/cmd/daemon/prune.go
@@ -86,7 +86,7 @@ func buildPruneCmd(parentCmd *cobra.Command) {
 	}
 }
 
-func pruningProgressBar(prunedCount, skippedCount, leftBlock, totalCount uint32) {
+func pruningProgressBar(prunedCount, skippedCount, leftBlocks, totalCount uint32) {
 	percentage := float64(prunedCount+skippedCount) / float64(totalCount) * 100
 	if percentage > 100 {
 		percentage = 100
@@ -96,6 +96,6 @@ func pruningProgressBar(prunedCount, skippedCount, leftBlock, totalCount uint32)
 	filledLength := int(float64(barLength) * percentage / 100)
 
 	bar := strings.Repeat("=", filledLength) + strings.Repeat(" ", barLength-filledLength)
-	fmt.Printf("\r [%s] %.0f%% Pruned: %d | Skipped: %d | Left Block: %d", //nolint
-		bar, percentage, prunedCount, skippedCount, leftBlock)
+	fmt.Printf("\r [%s] %.0f%% Pruned: %d | Skipped: %d | Left Blocks: %d", //nolint
+		bar, percentage, prunedCount, skippedCount, leftBlocks)
 }
--- a/cmd/daemon/prune.go
+++ b/cmd/daemon/prune.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/pactus-project/pactus/cmd"
+	"github.com/pactus-project/pactus/store"
+	"github.com/pactus-project/pactus/util"
+	"github.com/spf13/cobra"
+)
+
+func buildPruneCmd(parentCmd *cobra.Command) {
+	pruneCmd := &cobra.Command{
+		Use:   "prune",
+		Short: "prune old blocks and transactions from client",
+		Long: "The prune command optimizes blockchain storage by removing outdated blocks and transactions, " +
+			"freeing up disk space and enhancing client performance. customize pruning criteria via flags to " +
+			"manage storage effectively.",
+	}
+	parentCmd.AddCommand(pruneCmd)
+
+	workingDirOpt := pruneCmd.Flags().StringP("working-dir", "w", cmd.PactusDefaultHomeDir(),
+		"a path to the working directory of node files")
+
+	pruneCmd.Run = func(_ *cobra.Command, _ []string) {
+		workingDir, _ := filepath.Abs(*workingDirOpt)
+		if util.IsDirNotExistsOrEmpty(workingDir) {
+			cmd.PrintErrorMsgf("The working directory is not exists: %s", workingDir)
+
+			return
+		}
+
+		conf, _, err := cmd.MakeConfig(workingDir)
+		cmd.FatalErrorCheck(err)
+
+		cmd.PrintLine()
+		cmd.PrintWarnMsgf("Warning: This command removes all your blocks and transactions and changes " +
+			"your node to a prune node. You cannot revert to a full node.")
+		cmd.PrintLine()
+		confirmed := cmd.PromptConfirm("Do you want to continue")
+		if !confirmed {
+			return
+		}
+		cmd.PrintLine()
+
+		str, err := store.NewStore(conf.Store)
+		cmd.FatalErrorCheck(err)
+
+		prunedCount := uint32(0)
+		skippedCount := uint32(0)
+		totalCount := uint32(0)
+
+		interrupt := make(chan os.Signal, 1)
+		signal.Notify(interrupt, os.Interrupt, syscall.SIGTERM)
+
+		go func() {
+			<-interrupt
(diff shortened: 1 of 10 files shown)


Ja7ad:
@b00f I don't have idea to fix this lint error:

```shell
cmd/daemon/prune.go:99:2: use of `fmt.Printf` forbidden by pattern `^(fmt\.Print(|f|ln)|print|println)$` (forbidigo)
        fmt.Printf("\r [%s] %.0f%% Pruned: %d | Skipped: %d | Left Block: %d",
```

store/store.go
```diff
@@ -379,25 +388,66 @@ func (s *store) IsBanned(addr crypto.Address) bool {
...
+			continue
+		}
+
+		resultFunc(0, 1, i)
+	}
+
+	return nil
+}
+
 func (s *store) pruneBlock(blockHeight uint32) (bool, error) { //nolint
```
kehiy, line 426: You can remove `//nolint` from here.

store/store.go
```diff
@@ -379,25 +388,66 @@ func (s *store) IsBanned(addr crypto.Address) bool {
...
+	pruningHeight := cert.Height() - retentionBlocks
+
+	// start pruning blocks and transactions
+	for i := pruningHeight; i >= 1; i-- {
+		deleted, err := s.pruneBlock(i)
+		if err != nil {
+			return err
+		}
+
+		if deleted {
```
kehiy, line 414: If `deleted` is false we we had an error and we returned the function and this case won't be executed IMO.

store/config.go
```diff
@@ -54,5 +56,11 @@ func (conf *Config) BasicCheck() error {
 		}
 	}
 
+	if conf.RetentionDays < 10 {
+		return ConfigError{
+			Reason: "retention days too small than 10 days",
```
kehiy, line 61:
```suggestion
			Reason: "Retention days can't be less than 10 days",
```

> Also I think we can set 6 days as minimum and 10 as recomended.

cmd/daemon/prune.go
```diff
@@ -0,0 +1,101 @@
...
+
+			return
+		}
+
+		conf, _, err := cmd.MakeConfig(workingDir)
+		cmd.FatalErrorCheck(err)
+
+		cmd.PrintLine()
+		cmd.PrintWarnMsgf("Warning: This command removes all your blocks and transactions and changes " +
+			"your node to a prune node. You cannot revert to a full node.")
```
kehiy, line 43: Bassicly they can revert to a full node. we can't add this to warning I think.
They can simply download the full data and start again, or use a backup.

cmd/daemon/prune.go
```diff
@@ -0,0 +1,101 @@
...
+		cmd.PrintLine()
+		confirmed := cmd.PromptConfirm("Do you want to continue")
```
kehiy, line 45:
```suggestion
		confirmed := cmd.PromptConfirm("Do you want to continue?")
```

kehiy (changes requested):

store/store.go
```diff
@@ -379,25 +388,66 @@ func (s *store) IsBanned(addr crypto.Address) bool {
...
+
+	retentionBlocks := uint32(s.config.RetentionDays * totalBlockInDay)
+
+	if cert.Height() < retentionBlocks {
+		return nil
+	}
+
+	pruningHeight := cert.Height() - retentionBlocks
+
+	// start pruning blocks and transactions
```
kehiy, line 407: And the last thing, the logic of this loop is obvious and self explain, I think we can remove this comment.

store/store.go
Ja7ad: This skipped removed or not found block.

store/store.go
Ja7ad: This function needs to unit test.

kehiy: I know, I was talking about removing the comment...

store/store.go
kehiy: But when its false we have an error and we won't continue, how this block get executed at all?

Ja7ad: > But when its false we have an error and we won't continue, how this block get executed at all?

- false + nil = skipped non exists block
- true + nil = block pruned
- false + err = error and stop process

cmd/daemon/prune.go
```diff
@@ -0,0 +1,101 @@
...
+	pruneCmd := &cobra.Command{
+		Use:   "prune",
+		Short: "prune old blocks and transactions from client",
+		Long: "The prune command optimizes blockchain storage by removing outdated blocks and transactions, " +
+			"freeing up disk space and enhancing client performance. customize pruning criteria via flags to " +
+			"manage storage effectively.",
+	}
+	parentCmd.AddCommand(pruneCmd)
+
+	workingDirOpt := pruneCmd.Flags().StringP("working-dir", "w", cmd.PactusDefaultHomeDir(),
```
b00f, line 27: Can you make a function for the `working-dir` flag like:
```
func addWorkingDirOption(c *cobra.Command) *string {
	return c.Flags().StringP("working-dir", "w", cmd.PactusDefaultHomeDir(),
		"the path to the working directory that keeps the wallets and node files")
}

```

cmd/daemon/prune.go
```diff
@@ -0,0 +1,101 @@
...
+			cmd.PrintErrorMsgf("The working directory is not exists: %s", workingDir)
+
+			return
+		}
+
+		conf, _, err := cmd.MakeConfig(workingDir)
+		cmd.FatalErrorCheck(err)
+
+		cmd.PrintLine()
+		cmd.PrintWarnMsgf("Warning: This command removes all your blocks and transactions and changes " +
```
b00f, line 56: Can we change to this:

```
    cmd.PrintWarnMsgf("This command removes all the blocks and transactions up to %d days ago " +
    "and converts the node to prune mode.", conf.store.RetentionDays)
```

cmd/daemon/prune.go
```diff
@@ -0,0 +1,117 @@
...
+			pruningProgressBar(prunedCount, skippedCount, totalCount)
+		})
+		cmd.PrintLine()
+		cmd.FatalErrorCheck(err)
+
+		str.Close()
+		_ = fileLock.Unlock()
+
+		cmd.PrintLine()
+		cmd.PrintInfoMsgf("✅ Your node successfully pruned and changed to prune client.")
```
b00f, line 98:
```suggestion
		cmd.PrintInfoMsgf("✅ Your node successfully pruned and changed to prune mode.")
```

store/config.go
```diff
@@ -54,5 +56,11 @@ func (conf *Config) BasicCheck() error {
 		}
 	}
```
b00f, line 58: Add  a function for `RetentionBlocks`
```
func (conf *NodeConfig) RetentionBlocks int {
   return conf.RetentionDays * 8640
}
```

store/store.go
```diff
@@ -379,25 +388,65 @@ func (s *store) IsBanned(addr crypto.Address) bool {
...
 
+func (s *store) Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error {
+	cert := s.lastCertificate()
+
+	// at genesis block
+	if cert == nil {
+		return nil
+	}
+
+	retentionBlocks := uint32(s.config.RetentionDays * totalBlockInDay)
```
b00f, line 399: Use config function.

b00f (changes requested):

b00f (approved):

b00f (approved):

b00f:
@Ja7ad Thanks.

PR #1404, opened 2024-07-10
feat(state): prune block on commit

### LINKED ISSUE #1393: Pruning on New Block 1:1
Once a new block is committed, the following operations should be performed:

1. If the node is a Full Node, no action is needed.
2. If the node is a Pruned Node, call the PruneBlock function with the block number equal to LastBlockHeight - RetentionBlocks.

This process ensures that for each new block added, one old block will be removed, keeping the store blocks up to the RetentionBlocks number.

https://pips.pactus.org/PIPs/pip-27#pruning-on-new-block

kehiy:
## Description

This PR adds prune on a new block for pruned nodes.

## Related issue
- Fixes #1393

### DIFF
--- a/store/interface.go
+++ b/store/interface.go
@@ -98,6 +98,7 @@ type Reader interface {
 	TotalValidators() int32
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
+	IsPruned() bool
 }
 
 type Store interface {
--- a/store/interface.go
+++ b/store/interface.go
@@ -99,6 +99,7 @@ type Reader interface {
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
 	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
+	PruneOnCommit() error
 }
 
 type Store interface {
--- a/store/interface.go
+++ b/store/interface.go
@@ -99,7 +99,6 @@ type Reader interface {
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
 	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
-	PruneOnCommit() error
 }
 
 type Store interface {
--- a/store/interface.go
+++ b/store/interface.go
@@ -98,7 +98,6 @@ type Reader interface {
 	TotalValidators() int32
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
-	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
 }
 
 type Store interface {
@@ -107,6 +106,7 @@ type Store interface {
 	UpdateAccount(addr crypto.Address, acc *account.Account)
 	UpdateValidator(val *validator.Validator)
 	SaveBlock(blk *block.Block, cert *certificate.BlockCertificate)
+	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
 	WriteBatch() error
 	Close()
 }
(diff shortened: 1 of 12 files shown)


store/interface.go
```diff
@@ -99,6 +99,7 @@ type Reader interface {
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
 	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
+	PruneOnCommit() error
```
b00f, line 102: We don't need to tell the store to prune the block; the store should know when to prune or not. Let each module do its job

store/store.go
```diff
@@ -425,7 +434,30 @@ func (s *store) Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) er
 	return nil
 }
 
-func (s *store) pruneBlock(blockHeight uint32) (bool, error) { //nolint
+func (s *store) PruneOnCommit() error {
```
b00f, line 437: This logic should move to commit block function.

b00f (changes requested):

store/store.go
kehiy: But we have `RetentionBlocks` in the store, and `isPruned` can be determined properly in the store only.

Any way to have them in state as well without issues?

b00f: We can export a function, `IsPruned`, to inform the outside world that the store is in Prune mode. Remember, based on PIP-27, `pruneBlock` is private, so pruning occurs inside the store.

store/store.go
```diff
@@ -147,6 +155,24 @@ func (s *store) SaveBlock(blk *block.Block, cert *certificate.BlockCertificate)
...
+		if lc == nil {
+			return
+		}
+
+		pruneH := lc.Height() - s.config.RetentionBlocks()
+		if pruneH <= 0 {
+			return
+		}
+
+		deleted, err := s.pruneBlock(pruneH)
```
Ja7ad, line 170: After calling `pruneBlock` you need the call method `WriteBatch`

store/store.go
```diff
@@ -93,13 +94,20 @@ func NewStore(conf *Config) (Store, error) {
 		txStore:        newTxStore(db, conf.TxCacheSize),
 		accountStore:   newAccountStore(db, conf.AccountCacheSize),
 		validatorStore: newValidatorStore(db),
+		isPruned:       false,
```
Ja7ad, line 97: Don't require to set the default value, because the default value boolean is false

Ja7ad (changes requested):

store/store.go
b00f: @Ja7ad  It is called externally.

store/store.go
b00f: It's okay. Let it be. Some linters are sensitive about not setting values explicitly.

store/interface.go
```diff
@@ -107,6 +106,7 @@ type Store interface {
 	UpdateAccount(addr crypto.Address, acc *account.Account)
 	UpdateValidator(val *validator.Validator)
 	SaveBlock(blk *block.Block, cert *certificate.BlockCertificate)
+	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
```
b00f, line 109: Remove this function, and add `IsPrunned() bool`

b00f, line 109: Remove this function

kehiy: I just moved it no added it.

store/interface.go
```diff
@@ -98,7 +98,6 @@ type Reader interface {
 	TotalValidators() int32
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
-	Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) error
```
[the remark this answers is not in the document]
kehiy: Done

b00f (approved):

Ja7ad (approved):

PR #1406, opened 2024-07-11
test: add tests for pruning store

b00f:
## Description

This PR adds tests for pruning store and also refactor config tests.


store/store.go
```diff
@@ -441,15 +439,11 @@ func (s *store) Prune(resultFunc func(pruned, skipped, pruningHeight uint32)) er
...
-			resultFunc(1, 0, i)
-
-			continue
-		}
-
-		resultFunc(0, 1, i)
+		resultFunc(deleted, i)
 	}
 
+	s.isPruned = true
```
Ja7ad, line 445: This value stores in memory after prune needs to restart node.

Ja7ad (changes requested):

store/store.go
b00f: Done.

store/store.go
```diff
@@ -415,39 +419,37 @@ func (s *store) IsPruned() bool {
...
-		if err := s.WriteBatch(); err != nil {
+		if err := s.writeBatch(); err != nil {
 			return err
 		}
 
-		if deleted {
-			resultFunc(1, 0, i)
-
-			continue
+		if resultFunc(deleted, i) {
```
Ja7ad, line 449: Does not required return bool and check result func.

I think remove return bool.

store/store_test.go
```diff
@@ -23,8 +23,8 @@ type testData struct {
 func testConfig() *Config {
 	return &Config{
 		Path:               util.TempDirPath(),
-		TxCacheSize:        1024,
-		SortitionCacheSize: 1024,
+		TxCacheWindow:      1024,
```
Ja7ad, line 26: Old name is explicit for here, I think size better as window.

Ja7ad (changes requested):

store/store_test.go
b00f: This cache keeps transactions if they are not older than 8,640 blocks. We don't have a fixed size but a fixed window.
The old name was not accurate.

store/store.go
b00f: `resultFunc` is not a very good name here; it checks if the loop should continue or not. This prevents a random crash if the user presses Ctrl+C while Prune is working.

Ja7ad (approved):

PR #1420, opened 2024-07-16
feat(grpc): adding is pruned and pruning height in blockchain info API

### LINKED ISSUE #1418: Adding Prune information to BlockchainInfo API
Add two properties to BlockchainInfo API:
1- Is Pruned
2- PruningHeight

kehiy:
## Related issue

- Fixes #1418


state/state.go
```diff
@@ -762,3 +762,21 @@ func (st *state) AllPendingTxs() []*tx.Tx {
 
 	return st.txPool.AllPendingTxs()
 }
+
+func (st *state) IsPruned() bool {
+	st.lk.RLock()
+	defer st.lk.RUnlock()
```
Ja7ad, line 768: `IsPruned` is only set when the node is running and does not have any writes during application execution. Therefore, there is no need to lock it here.

store/store.go
```diff
@@ -487,3 +487,7 @@ func (s *store) pruneBlock(blockHeight uint32) (bool, error) {
 
 	return true, nil
 }
+
+func (s *store) RetentionBlocks() uint32 {
+	return s.config.RetentionDays
```
Ja7ad, line 492: This is retention days not block. please change to:

```go
s.config.RetentionBlocks()
```

Ja7ad (changes requested):

kehiy:
@Ja7ad Hi, change requests fixed, please review.

state/state.go
```diff
@@ -762,3 +762,18 @@ func (st *state) AllPendingTxs() []*tx.Tx {
 
 	return st.txPool.AllPendingTxs()
 }
+
+func (st *state) IsPruned() bool {
+	return st.store.IsPruned()
```
b00f, line 767: The Public Methods should be guarded by lock.

store/mock.go
```diff
@@ -297,3 +297,7 @@ func (*MockStore) Prune(_ func(_ bool, _ uint32) bool) error {
 func (*MockStore) IsPruned() bool {
 	return false
 }
+
+func (*MockStore) RetentionBlocks() uint32 {
```
b00f, line 301: Do we need it?

store/store.go
```diff
@@ -487,3 +487,7 @@ func (s *store) pruneBlock(blockHeight uint32) (bool, error) {
 
 	return true, nil
 }
+
+func (s *store) RetentionBlocks() uint32 {
```
b00f, line 491: do we need it?

www/grpc/proto/network.proto
```diff
@@ -71,10 +61,14 @@ message GetNodeInfoResponse {
   repeated string local_addrs = 8;
   // List of protocols supported by the node.
   repeated string protocols = 9;
+  // Type of node pruned or full.
+  bool is_pruned = 10;
```
b00f, line 65: To be backward compatible please don't change the field numbers.

www/grpc/proto/network.proto
```diff
@@ -120,3 +114,13 @@ message PeerInfo {
   // Completed download sessions with the peer.
   int32 completed_sessions = 20;
 }
+
+// Response message containing information about the node's connections.
```
b00f, line 118:
```suggestion
// ConnectionInfo contains information about the node's connections.
```

Ja7ad (approved):

b00f (commented):
To be compatible with Bitcoin RPC, let's add these information to the blockchain-info. Check here: https://developer.bitcoin.org/reference/rpc/getblockchaininfo.html

www/grpc/proto/network.proto
```diff
@@ -71,10 +61,14 @@ message GetNodeInfoResponse {
   repeated string local_addrs = 8;
   // List of protocols supported by the node.
   repeated string protocols = 9;
+  // Type of node pruned or full.
```
b00f, line 64:
```suggestion
  // If the blocks are subject to pruning.
```

www/grpc/proto/network.proto
```diff
@@ -71,10 +61,14 @@ message GetNodeInfoResponse {
...
+  bool is_pruned = 10;
+  // The height of pruning.
```
b00f, line 66:
```suggestion
  // Lowest-height block stored (only present if pruning is enabled)
```

b00f (changes requested):

state/state.go
kehiy: Check: https://github.com/pactus-project/pactus/pull/1420#discussion_r1680422259

store/store.go
kehiy: Yes, we can't access it from state otherwise.

www/grpc/proto/network.proto
kehiy: We didn't changed fields. we just used some unused numbers.

kehiy: There was no field for 10 and 11 before this.

kehiy:
@b00f Also, why we need to be compatible with Bitcoin RPC? We are not a fork or L2.

Also no bitcoin application uses Pactus. so whats the reason to be compatible?

store/interface.go
```diff
@@ -99,6 +99,7 @@ type Reader interface {
 	LastCertificate() *certificate.BlockCertificate
 	IsBanned(addr crypto.Address) bool
 	IsPruned() bool
+	RetentionBlocks() uint32
```
b00f, line 102: Why don't we export "Pruningheight" instead on "RetentionBlocks"?

www/grpc/proto/network.proto
```diff
@@ -71,10 +61,14 @@ message GetNodeInfoResponse {
   repeated string local_addrs = 8;
   // List of protocols supported by the node.
   repeated string protocols = 9;
+  // If the blocks are subject to pruning.
+  bool is_pruned = 10;
```
b00f, line 65: Please moe it to BlockchAainInfo API

b00f (changes requested):

b00f:
I think pruning information fits better in BlockchainInfo rather than NodeInfo. Regarding Bitcoin compatibility, many developers are familiar with Bitcoin, so we should create something that looks familiar and is better organized.

store/interface.go
kehiy: Because the internal function that we call have the same name. why change it?

kehiy:
@b00f Blokchain info contains the chain info which is shared, node info contains the info about one node which is not shared between all nodes.

store/interface.go
b00f: RetentionBlocks is an internal configuration. We better export something that has more general information.

www/grpc/proto/network.proto
```diff
@@ -71,10 +61,14 @@ message GetNodeInfoResponse {
...
+  // Lowest-height block stored (only present if pruning is enabled)
+  int32 pruning_height = 11;
   // Clock offset of the node.
-  double clock_offset = 13;
```
b00f, line 75: @kehiy don't change the old numbers. Keep them the same for backward compatibility. I know we changed them in the past, but now more users are using Pactus, and we can't ask them to upgrade their applications all the time.

kehiy: @b00f Fixed, now we didn't changes the numbers, only added some of them. anything must be backward compatible now.

kehiy:
>  Note: we decided that to move these info to blockchain info instead of node info since it will be mostly used there, using this info you can find you whether this node can provide info about old blocks or not. In this way caller can check this even if node info API is disabled.

> Note2: we moved pruning height logic to store from state.

b00f (approved):

PR #1429, opened 2024-07-21
fix(grpc): show negative pruning height when is pruned false

Ja7ad:
## Description

show negative pruning height when is pruned false 

image


kehiy (approved):
LGTM

www/grpc/blockchain.go
```diff
@@ -33,6 +33,11 @@ func (s *blockchainServer) GetBlockchainInfo(_ context.Context,
 		cv = append(cv, s.validatorToProto(v))
 	}
 
+	pruningHeight := uint32(0)
+	if s.state.IsPruned() {
+		pruningHeight = s.state.PruningHeight()
```
b00f, line 38: This logic should be handled inside store.

PR #1430, opened 2024-07-21
fix(store): pruning height returns zero when store is not in prune mode

b00f:
## Description

Updated the `PruningHeight` function to correctly return zero if the store is not in prune mode.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -432,6 +432,7 @@ func (s *store) PruningHeight() uint32 {
 
 	// TODO: it can be optimized (and safer?) by keeping the last block height in memory.
 	cert := s.lastCertificate()
+
 	return cert.Height() - s.config.RetentionBlocks()
 }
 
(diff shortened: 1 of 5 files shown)


store/store.go
```diff
@@ -415,11 +415,27 @@ func (s *store) IsBanned(addr crypto.Address) bool {
...
+func (s *store) PruningHeight() uint32 {
+	s.lk.RLock()
+	defer s.lk.RUnlock()
+
+	if !s.isPruned {
+		return 0
+	}
+
+	// TODO: it can be optimized (and safer?) by keeping the last block height in memory.
+	cert := s.lastCertificate()
```
kehiy, line 434: @b00f We can check if cert is nil here.

Ja7ad: This is not efficient for check, is prune one time set for prune state.

Ja7ad (approved):

store/store.go
kehiy: Thanks, got it.

kehiy (approved):

PR #1438, opened 2024-07-26
fix(consensus): handle query for decided proposal

b00f:
## Description

When the validators decide not to change the proposer and the proposal gets locked, the `QueryProposal` message can be responded to by all validators. 
This ensures that if the proposer fails, the proposal can still be broadcasted by other validators.


kehiy (approved):
LGTM

PR #1441, opened 2024-07-27
chore(genesis): set deprecated fee values to 0 for localnet

### LINKED ISSUE #1341: Set fee paramaetrs to zero in genesis doc
Once creating a new genesis file for the LocalNet the deprecated fee parameters set to default values, like:
```
      "fee_fraction": 0.0001,
      "minimum_fee": 1000,
      "maximum_fee": 1000000, 
```

These parameters should set to zero and mark as deprecated.

kehiy:
## Related issue

- Fixes #1341

### DIFF
--- a/genesis/testnet.json
+++ b/genesis/testnet.json
@@ -9,9 +9,9 @@
         "bond_interval": 360,
         "unbond_interval": 181440,
         "sortition_interval": 17,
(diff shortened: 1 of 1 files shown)


b00f:
Genesis file can't be modified!

PR #1485, opened 2024-08-24
fix(wallet, cmd): adding ed25519_account in help and set as default

kehiy:
## Description

The new address type was missed in the wallet cmd help command. Also ed25519 is faster and widely supported in different wallet's, so we can set it as default and new users will have ed25519 address.


cmd/wallet/address.go
```diff
@@ -74,7 +74,7 @@ func buildNewAddressCmd(parentCmd *cobra.Command) {
 	parentCmd.AddCommand(newAddressCmd)
 
 	addressType := newAddressCmd.Flags().String("type",
-		wallet.AddressTypeBLSAccount, "the type of address: bls_account or validator")
+		wallet.AddressTypeEd25519Account, "the type of address: bls_account, ed25519_account and validator")
```
akbariandev, line 77: I think it's not common way to use ed25519 as default. Actually it always need password and that's not user friendly. But as you said, it's faster and may convince us to use it as default. In total I'm not agree with this change.

kehiy: I agree. Passwords can be a good reason to consider them harder to use. I prefer to set BLS as default and Ed25519 as recommended. any ideas?

> By recommended I mean just adding a `recommended` tag in the help command.

b00f: Agree with @akbariandev We need to test it properly.

kehiy:
@b00f @akbariandev I reverted the default change. I tried to test the CLI and when I give an input as password, If it is not same as my wallet password it returns invalid password error.

I think it consider it as wallet password not a password provided for this address. or maybe I'm using this wrong?

b00f (approved):

b00f:
What is default now?

akbariandev:
> @b00f @akbariandev I reverted the default change. I tried to test the CLI and when I give an input as password, If it is not same as my wallet password it returns invalid password error.
> 
> I think it consider it as wallet password not a password provided for this address. or maybe I'm using this wrong?

@kehiy It must compare provided password with wallet password. Actually this feature can create new Ed25519 address by wallet password.

cmd/wallet/address.go
b00f: @kehiy revert the first part, just update the help string.

kehiy:
@akbariandev This only affects the default and description of command, not the way that it inputs the password from user. maybe I understand your comment wrong?

kehiy:
@b00f @akbariandev @akbariandev Any updates? should we merge or close this PR?

b00f:
@kehiy please add it but not make it as default now. Let's do it in the next release.

kehiy:
@b00f done.

b00f (approved):

Issue #1494, opened 2024-08-29
Add `info` Sub-Command to `pactus-wallet`


b00f:
We need to add a new `info` sub-command to the `pactus-wallet`. 
This sub-command should provide users with important information about their wallet. 
Specifically, it should print the following details:

- **Wallet Version:** Display the current version of the wallet.
- **Creation Time:** Show the creation time of the wallet.
- **Encrypted:** Indicate whether the wallet is encrypted or not.
- **Network:** Indicate the network the wallet is associated with.

PR #1496, opened 2024-09-02
feat(cmd): pactus-wallet add info commands

maxipaz:
## Description

Adding `info` commands for `pactus-wallet`

## Related issue(s)

- https://github.com/pactus-project/pactus/issues/1494

### DIFF
--- a/cmd/wallet/info.go
+++ b/cmd/wallet/info.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"time"
+
+	"github.com/pactus-project/pactus/cmd"
+	"github.com/spf13/cobra"
+)
+
+// buildInfoCmd builds all sub-commands related to the wallet information.
+func buildInfoCmd(parentCmd *cobra.Command) {
+	infoCmd := &cobra.Command{
+		Use:   "info",
+		Short: "retrieving the wallet information.",
+	}
+
+	parentCmd.AddCommand(infoCmd)
+
+	infoCmd.Run = func(_ *cobra.Command, _ []string) {
+		wlt, err := openWallet()
+		cmd.FatalErrorCheck(err)
+
+		cmd.PrintInfoMsgf("version: %d", wlt.Version())
+		cmd.PrintInfoMsgf("created at: %s", wlt.CreationTime().Format(time.RFC3339))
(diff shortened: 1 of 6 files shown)


wallet/wallet.go
```diff
@@ -502,3 +502,15 @@ func (w *Wallet) SignMessage(password, addr, msg string) (string, error) {
 
 	return prv.Sign([]byte(msg)).String(), nil
 }
+
+func (w *Wallet) GetVersion() int {
```
b00f, line 506: Just a note: In our codebase, we typically avoid using `Get` as a prefix for getter methods. Let's keep our naming consistent with this convention.

cmd/wallet/main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 	buildAllTransactionCmd(rootCmd)
 	buildAllAddrCmd(rootCmd)
 	buildAllHistoryCmd(rootCmd)
+	buildAllInfoCmd(rootCmd)
```
b00f, line 66:
```suggestion
	buildInfoCmd(rootCmd)
```

cmd/wallet/info.go
```diff
@@ -0,0 +1,90 @@
...
+
+// buildAllInfoCmd builds all sub-commands related to the wallet information.
+func buildAllInfoCmd(parentCmd *cobra.Command) {
+	infoCmd := &cobra.Command{
+		Use:   "info",
+		Short: "retrieving the wallet information.",
+	}
+
+	parentCmd.AddCommand(infoCmd)
+	buildWalletVersionInfoCmd(infoCmd)
```
b00f, line 18: I don't think we need subcommands here. It would be more convenient for the user to get all the wallet information at once.

b00f (changes requested):

wallet/wallet.go
maxipaz: Fixed. I also renamed `GetHistory` function that was a bit below this function

cmd/wallet/info.go
maxipaz: Done 🙂

b00f (approved):

wallet/wallet.go
b00f: Thanks

PR #1500, opened 2024-09-10
fix(grpc): merge raw transaction methods to one rpc method

### LINKED ISSUE #1480: Simplify Transaction APIs by Merging GetRawXTransaction Methods Using oneof in Protocol Buffers
### **Describe the problem you'd like to have solved**

The current Transaction APIs have separate RPC methods for retrieving raw details of different types of transactions (e.g., transfer, bond, unbond, withdraw). Each method requires a specific request and returns a response, leading to redundancy in the codebase. The existing structure complicates the API and increases maintenance overhead.

### **Describe the ideal solution**

The ideal solution is to simplify the Transaction APIs by merging all `GetRawXTransaction` methods into a single `GetRawTransaction` method. This unified method would handle different types of transactions using the `oneof` keyword in the Protocol Buffer definition.

The proposed changes are:

**Current API:**
```proto
// GetRawTransferTransaction retrieves raw details of a transfer transaction.
rpc GetRawTransferTransaction(GetRawTransferTransactionRequest)
    returns (GetRawTransactionResponse);

// GetRawBondTransaction retrieves raw details of a bond transaction.
rpc GetRawBondTransaction(GetRawBondTransactionRequest)
    returns (GetRawTransactionResponse);

// GetRawUnbondTransaction retrieves raw details of an unbond transaction.
rpc GetRawUnbondTransaction(GetRawUnbondTransactionRequest)
    returns (GetRawTransactionResponse);

// GetRawWithdrawTransaction retrieves raw details of a withdraw transaction.
rpc GetRawWithdrawTransaction(GetRawWithdrawTransactionRequest)
    returns (GetRawTransactionResponse);
```

**Proposed API:**
```proto
// GetRawTransaction retrieves raw bytes of a transaction.
rpc GetRawTransaction(GetRawTransactionRequest)
    returns (GetRawTransactionResponse);
```

In this approach, the `GetRawTransactionRequest` will use the `oneof` keyword to specify the type of transaction, thus handling different transaction types within a single API call. The old API methods will be marked as deprecated.

### **Alternatives and current work-around**

Currently, we handle each transaction type with its specific RPC method, which leads to duplication and complexity. Another alternative could be to create separate handler functions within the same service method, but this would still result in some redundancy.

### **Additional context**

Consolidating the transaction APIs using the `oneof` keyword in Protocol Buffers will simplify the API, reduce redundancy, and make the codebase easier to maintain. This change would also provide more flexibility for future transaction types without needing additional RPC methods. Marking the old APIs as deprecated ensures backward compatibility during the transition.

Ja7ad:
## Description

Simplify Transaction APIs by Merging GetRawXTransaction Methods Using oneof in Protocol Buffers

## Related issue(s)

- Fixes #1480


www/grpc/proto/transaction.proto
```diff
@@ -19,21 +19,8 @@ service Transaction {
   rpc BroadcastTransaction(BroadcastTransactionRequest)
       returns (BroadcastTransactionResponse);
 
-  // GetRawTransferTransaction retrieves raw details of a transfer transaction.
```
b00f, line 22: Please keep these methods, but mark them as deprecated

www/grpc/buf/grpc-gateway.config.yaml
```diff
@@ -44,17 +44,8 @@ http:
     - selector: pactus.Transaction.CalculateFee
       get: "/pactus/transaction/calculate_fee"
 
-    - selector: pactus.Transaction.GetRawTransferTransaction
```
b00f, line 47: We can remove them here

kehiy (approved):
LGTM.

b00f (approved):

Ja7ad:
> @Ja7ad Are you sure this won't break the current behaviour of JSON-RPC? Please check

I checked it.

www/grpc/proto/transaction.proto
```diff
@@ -84,6 +87,22 @@ message BroadcastTransactionResponse {
...
 
+// Request message for retrieving raw details of transaction.
+message GetRawTransactionRequest {
+  // The lock time for the transaction. If not set, defaults to the last block
+  // height.
+  uint32 lock_time = 1;
+  // A memo string for the transaction.
+  string memo = 2;
+
+  oneof transaction {
```
b00f, line 98: It is payload, not transaction

www/grpc/proto/transaction.proto
```diff
@@ -148,6 +167,50 @@ message GetRawWithdrawTransactionRequest {
   string memo = 6;
 }
 
+// Request message for retrieving raw details of a transfer transaction.
+message RawTransfer {
```
b00f, line 171: These are defined as Payload, please use Paylaod messages

Ja7ad:
@b00f Done

www/grpc/proto/transaction.proto
```diff
@@ -167,8 +167,8 @@ message GetRawWithdrawTransactionRequest {
   string memo = 6;
 }
 
-// Request message for retrieving raw details of a transfer transaction.
-message RawTransfer {
+// Payload message for retrieving raw details of a transfer transaction.
+message TransferPayload {
```
b00f, line 171: Why do we have two payload here: `TransferPayload` and `PayloadTransfer`? Why can't we use the old one?

Ja7ad: Because have duplicate memo and lock time fields.

b00f: Here is two structs:

```proto
// Payload message for retrieving raw details of a transfer transaction.
message TransferPayload {
  // The sender's account address.
  string sender = 1;
  // The receiver's account address.
  string receiver = 2;
  // The amount to be transferred, specified in NanoPAC. Must be greater than 0.
  int64 amount = 3;
  // The transaction fee in NanoPAC. If not set, it is set to the estimated fee.
  int64 fee = 4;
}

// Payload for a transfer transaction.
message PayloadTransfer {
  // The sender's address.
  string sender = 1;
  // The receiver's address.
  string receiver = 2;
  // The amount to be transferred in NanoPAC.
  int64 amount = 3;
}
```

The `TransferPayload` uses `TransferPayload`. The `TransferPayload` should NOT include `fee`.
if you remove the `fee` both struct will be the same.

Ja7ad:
@b00f Please check removed duplicate payload.

www/grpc/transaction.go
```diff
@@ -219,6 +258,96 @@ func (s *transactionServer) GetRawWithdrawTransaction(_ context.Context,
...
+
+	amt := amount.Amount(r.Transfer.Amount)
+	fee := s.getFee(r.Transfer.Fee, amt)
+
+	transferTx := tx.NewTransferTx(lockTime, sender, receiver, amt, fee, tx.WithMemo(memo))
+
+	return transferTx.Bytes()
+}
+
+func (s *transactionServer) handleRawBond(r *pactus.GetRawTransactionRequest_Bond,
```
b00f, line 283: Please avoid one letter name for variables. Follow the distance rule.

www/grpc/transaction.go
```diff
@@ -191,6 +229,7 @@ func (s *transactionServer) GetRawUnbondTransaction(_ context.Context,
 	}, nil
 }
 
+// Deprecated: GetRawWithdrawTransaction.
```
b00f, line 232: What does this comment want to say?

Ja7ad: This comment says this message, method or function is deprecated.

b00f:
I think we should find a way to use [tx-builder](https://github.com/pactus-project/pactus/blob/main/wallet/tx_builder.go) here.

www/grpc/proto/transaction.proto
```diff
@@ -19,19 +19,22 @@ service Transaction {
...
   rpc GetRawTransferTransaction(GetRawTransferTransactionRequest)
       returns (GetRawTransactionResponse);
 
-  // GetRawBondTransaction retrieves raw details of a bond transaction.
+  // Deprecated: GetRawBondTransaction retrieves raw details of a bond transaction.
   rpc GetRawBondTransaction(GetRawBondTransactionRequest)
       returns (GetRawTransactionResponse);
 
-  // GetRawUnbondTransaction retrieves raw details of an unbond transaction.
+  // Deprecated: GetRawUnbondTransaction retrieves raw details of an unbond transaction.
```
b00f, line 33: Please add comment here: `Use GetRawTransaction instead.`

www/grpc/proto/transaction.proto
```diff
@@ -84,7 +91,24 @@ message BroadcastTransactionResponse {
...
+message GetRawTransactionRequest {
+  // The lock time for the transaction. If not set, defaults to the last block height.
+  uint32 lock_time = 1;
+  // A memo string for the transaction.
+  string memo = 2;
+  // The fee for the transaction in NanoPAC.
+  int64 fee = 3;
+
+  oneof payload {
+    PayloadTransfer transfer = 4;
```
b00f, line 104: Add comment here, then it will update document

alidevjimmy:
@b00f do we want to revert all changes in this PR or just remove deprecates for that four methods?

b00f:
1. remove deprecates
2. remove `get_raw_transaction`

PR #1512, opened 2024-09-29
fix(cmd): add flag debug to enable pprof

Ja7ad:
## Description

Add flag debug to enable pprof

### DIFF
--- a/www/http/config.go
+++ b/www/http/config.go
@@ -3,7 +3,7 @@ package http
 type Config struct {
 	Enable         bool   `toml:"enable"`
 	Listen         string `toml:"listen"`
-	EnableDebugger bool   `toml:"-"`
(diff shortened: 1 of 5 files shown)


www/http/config.go
```diff
@@ -1,14 +1,16 @@
 package http
 
 type Config struct {
-	Enable bool   `toml:"enable"`
-	Listen string `toml:"listen"`
+	Enable         bool   `toml:"enable"`
+	Listen         string `toml:"listen"`
+	EnableDebugger bool   `toml:"-"`
```
b00f, line 6: Add a comment here for the private config:

```

// Private configs
EnableDebugger bool   `toml:"-"`
```

b00f (approved):

b00f (approved):

PR #1552, opened 2024-10-16
feat(sync): add metric to track the network activity

b00f:
## Description

This PR adds metric modules designed to track network activity more effectively. It allows us to monitor the number of bundles and the amount of data (in bytes) sent or received from each peer in the network.


sync/peerset/peer/metric/metric.go
```diff
@@ -0,0 +1,54 @@
...
+type Metric struct {
+	TotalInvalid    Counter
+	TotalSent       Counter
+	TotalReceived   Counter
+	MessageSent     map[message.Type]*Counter
+	MessageReceived map[message.Type]*Counter
+}
+
+func NewMetric() Metric {
+	return Metric{
```
Ja7ad, line 19: Used pointer receiver for methods, better return pointer.


```suggestion
	return *Metric{
```

www/grpc/network.go
```diff
@@ -106,23 +96,47 @@ func (s *networkServer) GetNetworkInfo(_ context.Context,
...
 		NetworkName:         s.net.Name(),
 		ConnectedPeersCount: uint32(len(peerInfos)),
 		ConnectedPeers:      peerInfos,
-		SentBytes:           sentBytes,
-		ReceivedBytes:       receivedBytes,
+		MetricInfo:          metricToProto(ps.Metric()),
 	}, nil
 }
+
+func metricToProto(m metric.Metric) *pactus.MetricInfo {
```
Ja7ad, line 107: I think instead return Metric object, use Get method + locker or `sync.Map` for read metric details.

Ja7ad (changes requested):

sync/peerset/peer/metric/metric.go
b00f: To avoid memory fragmentation in the peerset, it's better not to use pointers.

www/grpc/network.go
Ja7ad: Here read fields of `Metric` may data race happen.

Ja7ad: Yes here is a copy of the metric object.

PR #1554, opened 2024-10-17
feat(txpool): calculate consumption when committing a new block

### LINKED ISSUE #1542: Calculate consumption when committing a new block
To calculate consumption, we can define a map that associates an address with the total size of transactions stored in the blockchain over the last 8640 blocks. When a new block is committed, we update this map by iterating over all transactions in the block and increasing the value for each address by the size of its transactions. Simultaneously, the block from 8640 blocks ago would be retrieved, and the value for each address would be decreased by the size of its transactions.

https://pips.pactus.org/PIPs/pip-31#:~:text=Calculating%20Consumption

Ja7ad:
## Description

Calculate consumption when committing a new block

## Related issue(s)

- Fixes #1542


txpool/txpool.go
```diff
@@ -133,10 +140,57 @@ func (p *txPool) checkTx(trx *tx.Tx) error {
...
+	blk, err := committedBlock.ToBlock()
+	if err != nil {
+		return err
+	}
+
+	for _, trx := range blk.Transactions() {
+		if trx.IsTransferTx() || trx.IsBondTx() || trx.IsWithdrawTx() {
+			signer := trx.Payload().Signer()
+			if v, ok := p.consumptionMap[signer]; ok {
+				p.consumptionMap[signer] = v - uint32(trx.SerializeSize())
```
Ja7ad, line 185: @b00f I don't know on test `TestCalculatingConsumption` why value `p.consumptionMap` for specific signer don't update.

txpool/txpool_test.go
```diff
@@ -96,11 +102,36 @@ func TestAppendAndRemove(t *testing.T) {
 	// Appending the same transaction again, should not return any error
 	assert.NoError(t, td.pool.AppendTx(testTrx))
 
-	td.pool.RemoveTx(testTrx.ID())
+	td.pool.removeTx(testTrx.ID())
 	assert.False(t, td.pool.HasTx(testTrx.ID()), "Transaction should be removed")
 	assert.Nil(t, td.pool.PendingTx(testTrx.ID()))
 }
 
+func TestCalculatingConsumption(t *testing.T) {
```
Ja7ad, line 110: @b00f What case need for check consumption map? I need to more details for test cases.

state/state.go
```diff
@@ -325,7 +325,6 @@ func (st *state) ProposeBlock(valKey *bls.ValidatorKey, rewardAddr crypto.Addres
 		// Only one subsidy transaction per blk
 		if txs[i].IsSubsidyTx() {
 			st.logger.Error("found duplicated subsidy transaction", "tx", txs[i])
-			st.txPool.RemoveTx(txs[i].ID())
```
Ja7ad, line 328: @b00f Please double check for remove this line.

state/state.go
```diff
@@ -434,8 +434,8 @@ func (st *state) CommitBlock(blk *block.Block, cert *certificate.BlockCertificat
 	st.store.SaveBlock(blk, cert)
 
 	// Remove transactions from pool and update consumption
-	if err = st.txPool.HandleCommittedBlock(blk); err != nil {
-		return err
+	if errHandleCommittedBlk := st.txPool.HandleCommittedBlock(blk); errHandleCommittedBlk != nil {
```
b00f, line 437: Why such strange name for error?

b00f: Could you please move this code down after `store.WriteBatch()`.

txpool/config.go
```diff
@@ -5,8 +5,9 @@ import (
 )
 
 type Config struct {
-	MaxSize int        `toml:"max_size"`
-	Fee     *FeeConfig `toml:"fee"`
+	MaxSize           int        `toml:"max_size"`
+	ConsumptionBlocks uint32     `toml:"-"`
```
b00f, line 9: Let's put the private field after public configs, and add a comment, like: `// Private configs`

txpool/config.go
```diff
@@ -17,8 +18,9 @@ type FeeConfig struct {
 
 func DefaultConfig() *Config {
 	return &Config{
-		MaxSize: 1000,
-		Fee:     DefaultFeeConfig(),
+		MaxSize:           1000,
+		ConsumptionBlocks: 8640,
```
b00f, line 22: Same here. move it down, after public configs.

txpool/txpool.go
```diff
@@ -19,14 +21,16 @@ import (
...
+	sandbox        sandbox.Sandbox
+	pools          map[payload.Type]pool
+	consumptionMap map[crypto.Address]uint32
+	broadcastCh    chan message.Message
+	strReader      store.Reader
+	logger         *logger.SubLogger
 }
 
-func NewTxPool(conf *Config, broadcastCh chan message.Message) TxPool {
+func NewTxPool(conf *Config, broadcastCh chan message.Message, storeReader store.Reader) TxPool {
```
b00f, line 33: Let put channels as last parameter. Parameters are set based on how important they are.

```go
func NewTxPool(conf *Config, storeReader store.Reader, broadcastCh chan message.Message ) TxPool {
```

txpool/txpool.go
```diff
@@ -133,10 +139,67 @@ func (p *txPool) checkTx(trx *tx.Tx) error {
...
+	}
+
+	for _, trx := range blk.Transactions() {
+		if trx.IsTransferTx() || trx.IsBondTx() || trx.IsWithdrawTx() {
+			signer := trx.Payload().Signer()
+			if v, ok := p.consumptionMap[signer]; ok {
+				// Decrease the consumption by the size of the transaction
+				v -= uint32(trx.SerializeSize())
+
+				// If the new value is zero, remove the signer from the consumptionMap
```
b00f, line 188: Move comment into the curve bracket { }.

b00f (changes requested):

txpool/config.go
akbariandev: Is it necessary to have ConsumptionBlocks configured? If the value is constantly equal to 8640 so it would be global constant and not a public field in config

txpool/txpool.go
```diff
@@ -133,10 +139,67 @@ func (p *txPool) checkTx(trx *tx.Tx) error {
...
+
+	return p.handleDecreaseConsumption(blk.Height())
+}
+
+func (p *txPool) handleIncreaseConsumption(trx *tx.Tx) {
+	if trx.IsTransferTx() || trx.IsBondTx() || trx.IsWithdrawTx() {
+		signer := trx.Payload().Signer()
+
+		p.consumptionMap[signer] += uint32(trx.SerializeSize())
+	}
```
alidevjimmy, line 160: I think it's better to have `atomic` increase to prevent race conditions on `consumptionMap`

Ja7ad: Here don't use unsafe memory, no need to use atomic.

Ja7ad: For the consumptionMap map[crypto.Address]uint32, atomic operations are typically not necessary unless multiple goroutines concurrently read and modify the same uint32 values in the map. A map in Go is not thread-safe, so accessing or modifying it concurrently requires synchronization (e.g., using sync.Mutex).

However, if the map is safely guarded by a lock or accessed in a single-threaded context, atomic operations for the values aren't required.

alidevjimmy: I mean, we can't guarantee scope of use of `consumptionMap` in long term.
Although we can guarantee that `handleIncreaseConsumption` is locked but it is likely that another function or goroutine accesses `consumptionMap` in different scope using different lock, which is not safe.

Ja7ad: `handleIncreaseConsumption` and `handleDecreaseConsumption` have parent lock from `HandleCommittedBlock`.

In HandleCommittedBlock we don't have concurrent for increase and decrease.

b00f (approved):

Issue #1546, opened 2024-10-11
Display fee configuration in node information


Ja7ad:
Display fee configuration from `config.toml` in node information.

```shell
/pactus/network/get_node_info
```

```json
{
  "moniker": "foobar",
...
  "fee": {
    "daily_limit": 250, // uint32
    "unit_price": 0.000005, // float64
    "fixed_fee": 0.01 // float64
  }
}
```

Ja7ad:
@b00f We display consumption configuration or not planning?

b00f:
I don't think we need to add in in gRPC. Please keep this issue open. We may add it in future.

Ja7ad:
> I don't think we need to add in in gRPC. Please keep this issue open. We may add it in future.

Ok, remove this issue from parent issue.

PR #1561, opened 2024-10-20
feat(grpc): add fee config to get node info api response

alidevjimmy:
## Description

This PR adds a node fee config to `GetNodeInfoResponse` of `GetNodeInfo` method.

## Related issue(s)

- #1546


node/node.go
```diff
@@ -91,7 +91,7 @@ func NewNode(genDoc *genesis.Genesis, conf *config.Config,
 	if conf.GRPC.BasicAuth != "" {
 		enableHTTPAuth = true
 	}
-	grpcServer := grpc.NewServer(conf.GRPC, st, syn, net, consMgr, walletMgr)
+	grpcServer := grpc.NewServer(conf.GRPC, st, syn, net, consMgr, walletMgr, conf.TxPool)
```
Ja7ad, line 94: @b00f What you think we pass just `conf` instead part of configuration?

Ja7ad:
@alidevjimmy In http server you need add fee configuration to node handler.

https://github.com/pactus-project/pactus/blob/80ea4db1644d6e552bcf216541db9a1b63512f6a/www/http/network.go#L80-L118

www/grpc/proto/network.proto
```diff
@@ -129,3 +131,13 @@ message CounterInfo {
   // Total number of bundles.
   uint64 Bundles = 2;
 }
+
+// FeeConfig holds the fee configuration that stored in node config.
+message FeeConfig {
+  // Fixed fee for each transaction.
+  double fixed_fee = 1;
+  // Number of bytes an account can send each day without paying a fee.
```
b00f, line 139: Please add this comment here: `The `daily_limit` is part of he consumptional fee model.`

www/grpc/proto/network.proto
```diff
@@ -129,3 +131,13 @@ message CounterInfo {
...
+  uint32 daily_limit = 2;
+  // Fee per byte in PAC.
```
b00f, line 141: Please add this comment here: `The `unit_price` is part of he consumptional fee model.`

b00f (changes requested):

node/node.go
b00f: Nope! We can't add parameter that simple!
Try to find another way...

b00f: Let's think out of box; gRPC almost have everything. Right?
Why don't we simply give him Node.

Ja7ad: We can give access to configuration via node?

alidevjimmy: node itself depends on gRPC

b00f: Node has all modules...

alidevjimmy:
> @alidevjimmy In http server you need add fee configuration to node handler.
> 
> https://github.com/pactus-project/pactus/blob/80ea4db1644d6e552bcf216541db9a1b63512f6a/www/http/network.go#L80-L118

It's Done.

node/node.go
alidevjimmy: If we pass node to gRPC we will encounter **cycle import**.

b00f:
Thanks @alidevjimmy  but This PR only adds complexity to the code, without any clear reason. Let's close it.
In case we really need this information, we can think about it.

PR #1567, opened 2024-10-21
fix(consensus): send decided vote for previous round on query vote

b00f:
## Description

Upon receiving a QueryVote, if there is any decided vote for the last round, it should be broadcasted. Previously, it broadcast the decided vote for the query round. This helps validators who are behind the network to sync faster.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -514,13 +514,13 @@ func (cs *consensus) HandleQueryVote(height uint32, round int16) *vote.Vote {
 	votes := []*vote.Vote{}
 	switch {
 	case round < cs.round:
-		// Past round: Only broadcast cp:decided votes
-		vs := cs.log.CPDecidedVoteSet(round)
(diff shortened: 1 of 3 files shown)


consensus/consensus.go
```diff
@@ -514,13 +514,13 @@ func (cs *consensus) HandleQueryVote(height uint32, round int16) *vote.Vote {
 	votes := []*vote.Vote{}
 	switch {
 	case round < cs.round:
-		// Past round: Only broadcast cp:decided votes
-		vs := cs.log.CPDecidedVoteSet(round)
+		// Past round: broadcast cp:decided votes for the previous round.
+		vs := cs.log.CPDecidedVoteSet(cs.round - 1)
```
Ja7ad, line 518: @b00f can you more explain it, why broadcast for the previous round?

b00f: Please check the comment again...

ja7ad (approved):

PR #1577, opened 2024-11-01
fix(grpc): set Bond public key for decoded transaction

### LINKED ISSUE #1576: GRPC: Bond Transaction doesn't decode properly if it has public key
## Description

The Bond Transactions with a public key do not decode properly in the gRPC endpoint for the `Transaction_GetTransaction` API. The final output has no Public Key. For example, you can check the details for this transaction ID: `d194b445642a04ec78ced4448696e50b733f2f0b517a23871882c0eefaf1c28f`.

### How to Reproduce

Call the `Transaction_GetTransaction` API with the transaction ID provided above.

### Expected Behavior

The API should verify if the transaction includes a public key and, if so, include it within the proto object.

b00f:
## Description

This PR verifies whether the Bond transaction includes the public key and, if so, sets it for the proto payload. Previously, the Validator public key was missed even when the transaction has one.

## Related issue(s)

- Fixes #1576

### DIFF
--- a/types/tx/tx_test.go
+++ b/types/tx/tx_test.go
@@ -94,7 +94,7 @@ func TestBasicCheck(t *testing.T) {
 
 	t.Run("LockTime is not defined", func(t *testing.T) {
 		trx := tx.NewTransferTx(0,
-			ts.RandAccAddress(), ts.RandAccAddress(), ts.RandAmount(), ts.RandAmount())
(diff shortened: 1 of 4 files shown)


Ja7ad:
@b00f Please fix the lint issue

ja7ad (approved):

PR #1589, opened 2024-11-09
fix(txpool, cmd, gtk): broadcast transactions with zero fee

b00f:
## Description

This PR ensures that transactions with zero fees can be broadcasted to the network, even if the tx-pool rejects them and the configuration does not accept zero-fee transactions. 
This way, users can broadcast their transactions with zero fees, but their nodes will still drop them if they are deemed unacceptable.

This PR also refactored the `txpool` tests and simplify them.


txpool/txpool_test.go
```diff
@@ -92,119 +94,204 @@ func (td *testData) shouldPublishTransaction(t *testing.T, txID tx.ID) {
...
-		})
-		td.store.SaveBlock(blk, cert)
+		blk, cert := td.GenerateTestBlock(tt.height, testsuite.BlockWithTransactions(tt.txs))
+		td.sbx.TestStore.SaveBlock(blk, cert)
 
 		// Handle the block in the transaction pool
 		td.pool.HandleCommittedBlock(blk)
+
+		// Uncomment me to track consumption map.
+		// fmt.Println(td.pool.consumptionMap)
```
Ja7ad:
```suggestion
t.Log(td.pool.consumptionMap)
```

ja7ad (changes requested):
@b00f I think better check `daily_limit <= 0` in `BasicCheck`

```suggestion
	if conf.Fee.DailyLimit <= 0 {
		return ConfigError{
			Reason: "dailyLimit can't be zero",
		}
	}
```

ja7ad (approved):

PR #1594, opened 2024-11-11
refactor(cmd): get first account address from wallet as reward address

b00f:
## Description

This PR introduces a change in behavior for obtaining the reward address.
If the reward address is not specified in the config, it attempts to retrieve it from the wallet.
The first Ed25519 account address is considered as the reward address, and if not available, the first BLS account address is used.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -296,21 +296,21 @@ func CreateNode(numValidators int, chain genesis.ChainType, workingDir string,
 ) ([]string, string, error) {
 	// To make process faster, we update the password after creating the addresses
 	walletPath := PactusDefaultWalletPath(workingDir)
-	walletInstance, err := wallet.Create(walletPath, mnemonic, "", chain)
+	wlt, err := wallet.Create(walletPath, mnemonic, "", chain)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cmd/cmd.go
```diff
@@ -547,61 +547,63 @@ func RecoverConfig(confPath string, defConf *config.Config, chainType genesis.Ch
...
-	}
 
-	// Check if reward addresses are account address
-	for _, addr := range rewardAddrs {
-		if !addr.IsAccountAddress() {
-			return nil, fmt.Errorf("reward address is not an account address: %s", addr)
+	// Case 2: One reward address is specified in the config file.
+	case len(confRewardAddrs) == 1:
+		// Use this single address for all validators.
+		addr, _ := crypto.AddressFromString(confRewardAddrs[0])
```
akbariandev, line 592: Why are these kinds of errors never handled? This does not make an impact here but maybe has affect in transfer reward process when this address is not valid.

b00f: In BasicCheck, we verify the validity of the config file, so by this point, we know the config is valid.

ja7ad (approved):

PR #1608, opened 2024-11-22
fix(grpc): define address for the aggregated public key

kehiy:
## Description

when we merge public keys, we can get the resulting address.


b00f (approved):
@alidevjimmy please review

Ja7ad (approved):

PR #1646, opened 2024-12-18
feat(sync): propagationPolicy added to handle expired message

akbariandev:
## Description

By this PR before the propagation of each message, the expiration of the message will be checked through some policies. If the message was expired, it will droped.


b00f:
Unacceptable

PR #1653, opened 2024-12-22
feat(cmd): read password from file

### LINKED ISSUE #1610: Read wallet password from the file

Currently, to start the node, the user must provide the wallet password as an argument to the `pactus-daemon` executable. This approach is particularly vulnerable, even if `pactus-daemon` is run as `sudo` by a non-root user. For example, if the server gets compromised, it is easy to retrieve the wallet password without needing the `sudo` password by using a command like:

```
ps -p $(pidof pactus-daemon) -o args
```

To improve the security of the node, we can store the password in a file that is readable only by `sudo`. Then, the `pactus-daemon` process can be run with `sudo` and read the password from this file. This approach enhances security by protecting the wallet password with the `sudo` password.

Ja7ad:
## Description

Add switch `password-from-file` to read password from file content.

## Related issue(s)

- Fixes #1610


util/utils.go
```diff
@@ -160,3 +163,23 @@ func FormatBytesToHumanReadable(bytes uint64) string {
 
 	return fmt.Sprintf("%.2f %s", value, unit)
 }
+
+// ReadFileContent reads the content of the file at the given path
+// up to the specified maxSize. It returns the content as a string
+// and any error encountered.
+func ReadFileContent(filePath string) (string, error) {
```
b00f, line 170: We have `util.ReadFile`, use that function instead.

b00f (changes requested):

cmd/daemon/start.go
```diff
@@ -49,8 +54,15 @@ func buildStartCmd(parentCmd *cobra.Command) {
 			}
 
 			var password string
+
 			if *passwordOpt != "" {
 				password = *passwordOpt
+			} else if *passwordFromFileOpt != "" {
+				b, err := util.ReadFile(*passwordFromFileOpt)
+				if err != nil {
+					return "", false
```
b00f, line 63: What happens here? Do we show user the error?

Ja7ad: > What happens here? Do we show user the error? 

We show invalid password, don't need to handle error.

b00f: No, better we show proper error here, file error and exit

cmd/daemon/start.go
```diff
@@ -49,8 +54,13 @@ func buildStartCmd(parentCmd *cobra.Command) {
 			}
 
 			var password string
+
 			if *passwordOpt != "" {
 				password = *passwordOpt
+			} else if *passwordFromFileOpt != "" {
+				b, err := util.ReadFile(*passwordFromFileOpt)
+				cmd.FatalErrorCheck(err)
+				password = strings.TrimSpace(string(b))
```
b00f, line 63:
```suggestion

				password = strings.TrimSpace(string(b))
```

b00f (approved):

b00f (approved):

PR #1666, opened 2025-01-13
feat(zeromq): add block info publisher

### LINKED ISSUE #1658: PIP-36: Implement Block Info Publisher
## Description:

Implement the `Raw Block Publisher`, which will handle notifications related to raw block data. 
This publisher should implement the `Publisher` interface and handle sending raw block data as per the PIP-36 specifications.  

### Tasks:

- Implement `BlockInfoPublisher` struct under `www/zmq`  
- Ensure it correctly implements the `Publisher` interface  
- Add functionality for sending raw block data notifications  
- Write unit tests to ensure the publisher behaves as expected

Ja7ad:
## Description

Add block info publisher 

## Related issue(s)

- Fixes #1658


state/state.go
```diff
@@ -752,5 +752,8 @@ func (st *state) publishEvent(msg any) {
 		return
 	}
 
-	st.eventCh <- msg
+	select {
```
b00f, line 755: Why do we need select here?

www/zmq/block_info_publisher.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 type blockInfoPub struct {
+	seqNo uint32
 	basePub
```
b00f, line 11: Move it one line up. Add one new line between them

www/zmq/util.go
```diff
@@ -0,0 +1,30 @@
+package zmq
+
+import (
+	"encoding/binary"
+
+	"github.com/pactus-project/pactus/crypto"
+)
+
+func makeTopicMsg(parts ...any) []byte {
```
b00f, line 9: I feel we can add it to the base and then add topic and seq at the beginning and end of all topic messages.

www/zmq/block_info_publisher_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+	"time"
+
+	"github.com/go-zeromq/zmq4"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBlockInfoPublisher(t *testing.T) {
+	td := setup(t)
+	defer td.cleanup()
```
Ja7ad, line 16: @b00f We have race condition here, what you think to fix it?

state/state.go
Ja7ad: > Why do we need select here?

Select prevent panic on send message on closed channel.

b00f (approved):

PR #1668, opened 2025-01-16
fix(util): move dir for disk on different mount

### LINKED ISSUE #1579: Failed import prune data to mounted path
## Description

When import prune data to mounted path got error `invalid cross-device link`

The error EXDEV (18): invalid cross-device link occurs because os.Rename cannot move files or directories across different file systems (e.g., from one mounted drive to another). os.Rename works only when the source and destination are on the same filesystem.

### What Happened

```shell
✔ snapshot 2024-11-01 (157.70 MB)

   0% [                              ]  data.zip (0.00 Bytes/157.70 MB) 2024/11/01 11:02:57 download completed


Extracting files...
Moving data...
failed to move directory from /tmp/pactus689140844/data to /home/javad/Work/Pactus/data/testnet/data: rename /tmp/pactus689140844/data /home/javad/Work/Pactus/data/testnet/data: invalid cross-device link
```

Error happen in:

https://github.com/pactus-project/pactus/blob/9ddfd957b038220888fa9213286f5df9a85bda49/util/io.go#L217-L219

BuidlWithShivam:
## Description

os.Rename gives error when moving cross disk. Instead, we use io.copy to copy from source to destination and delete source.

## Related issue(s)

- Fixes #1579


Ja7ad:
@BuidlWithShivam Possible write unit test for this functionality in temp path?

b00f:
@Ja7ad if you confirm this is working fine, we can merge it

Ja7ad:
@b00f @BuidlWithShivam I got error on import

```shell
failed to move directory from /tmp/pactus2236639250/data to /home/javad/test/data: failed to copy /tmp/pactus2236639250/data to /home/javad/test/data : write /home/javad/test/data: copy_file_range: is a directory
```

PR #1682, opened 2025-02-02
fix(firewall): remove decoding bundle for expired message

### LINKED ISSUE #1649: Remove decoding bundle for expired message
We decode the bundle to detect the consensus height for nodes running on version 1.6.3 and lower. This helps maintain backward compatibility until the majority of the network upgrades. We can remove the bundle decoding in version 1.8.0 to simplify the code logic.

ja7ad:
## Description

Remove decoding bundle for expired message

## Related issue(s)

- Fixes #1649


b00f (approved):

sync/firewall/firewall.go
```diff
@@ -217,35 +217,20 @@ func (f *Firewall) isExpiredMessage(msgData []byte) bool {
...
 
-	// TODO: we remove on the version v1.8.0
-	consensusHeightExtracted := false
-	var consensusHeight uint32
+	// Validate the format before extracting consensusHeight
 	consensusHeightBytes := msgData[msgLen-6:]
-	// Check if consensus height is set. Refer to the bundle encoding for more details.
-	if consensusHeightBytes[0] == 0x04 && consensusHeightBytes[1] == 0x1a {
-		consensusHeight = binary.BigEndian.Uint32(consensusHeightBytes[2:])
+	if consensusHeightBytes[0] != 0x04 || consensusHeightBytes[1] != 0x1a {
```
b00f, line 222: This line can be removed.

b00f (approved):

b00f (approved):

b00f (changes requested):
@Ja7ad 

If the reported consensus height in a message (here)  is not same as the actual consensus height of a message, sender should be banned.

Ja7ad:
> @Ja7ad
> 
> If the reported consensus height in a message (here) is not same as the actual consensus height of a message, sender should be banned.

What your suggestion for fix this issue?

b00f (approved):

PR #1683, opened 2025-02-02
feat(wallet-cmd): add Neuter command for wallet

### LINKED ISSUE #1661: Add `Neuter` command for wallet
## Describe the problem you'd like to solve

Currently, there is no way to *neuter* a wallet.  
A neutered wallet is a wallet that does not store private keys or seeds. It is a read-only wallet and can only be used to retrieve balances or stakes.  
For BLS keys, it can also create new addresses without needing the private key.  

While the vault object supports neutered wallets, we currently lack a command to convert a full wallet into a neutered wallet.

We need to add a new command to the wallet application to neuter a wallet.  
The command should check if the wallet is full or already neutered. If it is a full wallet, it should create a neutered copy and save it to the specified path.

Ja7ad:
## Description

Add Neuter command for wallet

## Related issue(s)

- Fixes #1661

### DIFF
--- a/cmd/wallet/neuter.go
+++ b/cmd/wallet/neuter.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"github.com/pactus-project/pactus/cmd"
+	"github.com/spf13/cobra"
+)
+
+func buildNeuterCmd(parentCmd *cobra.Command) {
+	neuterCmd := &cobra.Command{
+		Use:   "neuter",
+		Short: "Convert full wallet to read-only wallet and can only be used to retrieve balances or stakes",
+	}
+	parentCmd.AddCommand(neuterCmd)
+
+	neuterCmd.Run = func(_ *cobra.Command, _ []string) {
+		wlt, err := openWallet()
+		cmd.FatalErrorCheck(err)
+
+		neuteredWallet := wlt.Neuter()
+
+		err = neuteredWallet.Save()
+		cmd.FatalErrorCheck(err)
+	}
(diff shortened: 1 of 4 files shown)


cmd/wallet/neuter.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/pactus-project/pactus/cmd"
+	"github.com/spf13/cobra"
+)
+
+func buildNeuterCmd(parentCmd *cobra.Command) {
+	neuterCmd := &cobra.Command{
+		Use:   "neuter",
+		Short: "convert full wallet to read-only wallet and can only be used to retrieve balances or stakes",
```
b00f, line 11:
```suggestion
		Short: "convert full wallet to neutered wallet and can only be used to retrieve balances or stakes",
```

wallet/wallet.go
```diff
@@ -548,3 +548,16 @@ func (w *Wallet) Info() *Info {
...
 	}
 }
+
+func (w *Wallet) Neuter() *Wallet {
+	clonedStore := w.store.Clone()
+	clonedStore.Vault = w.store.Vault.Neuter()
+
+	neuteredWallet := &Wallet{
+		store:      clonedStore,
+		path:       w.path + "_neutered",
```
b00f, line 558:
```suggestion
		path:       w.path + ".neutered",
```

b00f (approved):

b00f (approved):

PR #1703, opened 2025-03-09
fix(wallet): set Argon2 derived bytes for AES IV

b00f:
## Description

This PR updates the encryption logic to use Argon2-derived bytes for the Initialization Vector (IV) in AES, replacing the legacy approach of reusing the salt as the IV. 

### Explanation:

In the legacy or previous approach, we used the same salt in the Password Hasher (Argon2 in this case) as the Initialization Vector (IV) for the AES encryption algorithm. The seed is random and should not be an issue, but some documentation recommends never reusing a seed, even for different purposes.  

To address this, we propose extending the key length from 32 bytes to 48 bytes, using the first 32 bytes as the encryption key and the remaining 16 bytes as the IV.  

In OpenSSL, the salt is public. I found [this code](https://github.com/openssl/openssl/blob/91c6e157c696e8fee7320408ddb959ecf233fbaf/apps/enc.c#L546).
The encrypted files usually start with `Salted__`. 

Based on [this](https://docs.openssl.org/3.3/man1/openssl-enc/#options) documentation, the IV is derived from the password in OpenSSL.  Check the `-iv` section:

> The actual IV to use: this must be represented as a string comprised only of hex digits. When only the key is specified using the -K option, the IV must explicitly be defined. When a password is being specified using one of the other options, the IV is generated from this password.

We follow the same approach in this PR. [Here](https://github.com/openssl/openssl/blob/91c6e157c696e8fee7320408ddb959ecf233fbaf/apps/enc.c#L585) I found the code that generates IV in OpenSSL.

### DIFF
--- a/wallet/encrypter/encrypter.go
+++ b/wallet/encrypter/encrypter.go
@@ -91,7 +91,6 @@ func NopeEncrypter() Encrypter {
 // The default encrypter uses Argon2ID as password hasher and AES_256_CBC as
 // encryption algorithm.
 func DefaultEncrypter(opts ...Option) Encrypter {
-
 	argon2dParameters := &argon2dParameters{
 		iterations:  defaultIterations,
 		memory:      defaultMemory,
@@ -238,25 +237,25 @@ func (e *Encrypter) Decrypt(cipherText, password string) (string, error) {
 		// Encrypter method
 		switch funcs[1] {
 		case nameFuncAES256CTR:
-			var iv, cipherKey []byte
+			var initVec, cipherKey []byte
 
 			switch keyLen {
 			case 0:
 				// This case supports legacy encryption methods where the same salt is reused as the IV.
 				cipherKey = derivedByte
-				iv = salt
+				initVec = salt
 
 			case 48:
 				// The first 32 bytes are used as the encryption key, and the last 16 bytes are used as the IV.
 				cipherKey = derivedByte[:32]
-				iv = derivedByte[32:]
+				initVec = derivedByte[32:]
 
 			default:
 				return "", ErrInvalidParam
 			}
 
 			enc := data[16 : len(data)-4]
-			text = string(aesCTRCrypt(enc, iv, cipherKey))
+			text = string(aesCTRCrypt(enc, initVec, cipherKey))
 
 			// MAC method
 			switch funcs[2] {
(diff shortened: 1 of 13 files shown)


wallet/encrypter/encrypter.go
```diff
@@ -47,11 +48,20 @@ const (
...
 	nameFuncNope      = ""
 	nameFuncArgon2ID  = "ARGON2ID"
 	nameFuncAES256CTR = "AES_256_CTR"
+	nameFuncAES256CBC = "AES_256_CBC"
 	nameFuncMACv1     = "MACV1"
+
+	// Parameter Choice
+	// https://www.rfc-editor.org/rfc/rfc9106.html#section-4
+	defaultIterations  = 3
+	defaultMemory      = 65536 // 2 ^ 16
```
Ja7ad, line 62: This method improves readability while keeping the memory definition precise.

```suggestion
	defaultMemory      = 64 << 20 // 2 ^ 16 = 64 MB
```

wallet/encrypter/encrypter.go
```diff
@@ -47,11 +48,20 @@ const (
...
+	defaultParallelism = 4
+	defaultKeyLen      = 48
```
Ja7ad, line 64: I can't find recommended value for `KeyLen` and what is it keylen?

wallet/encrypter/encrypter_test.go
```diff
@@ -66,3 +67,31 @@ func TestEncrypter(t *testing.T) {
 	_, err = enc.Decrypt(cipher, "invalid-password")
 	assert.ErrorIs(t, err, ErrInvalidPassword)
 }
+
+func TestEncrypterV3(t *testing.T) {
+	enc := &Encrypter{
```
Ja7ad, line 72: I think you can define two constructor functions for versioning to improve readability. For example:

```go
func NewEncryptorV2() *Encryptor {}
```

```go
func NewEncryptorV3() *Encryptor {}
```

ja7ad (changes requested):

wallet/encrypter/encrypter.go
b00f: I understand 2 << 16 or (65536) better than 64 << 20!

wallet/encrypter/encrypter_test.go
b00f: No, We only have one constructor that is `DefaultEncryptor`. 
the rest are initialized by JSON unmarshaling.

wallet/encrypter/encrypter.go
Ja7ad: Ok change to `2 << 16` it's better.

wallet/encrypter/encrypter.go
b00f: Check above explanation

ja7ad (approved):

ja7ad (approved):

PR #1770, opened 2025-05-02
feat(rest): define base path for REST APIs

b00f:
## Description

This PR defines the base_path for the REST APIs and allows users to customize it.

### DIFF
--- a/www/rest/config.go
+++ b/www/rest/config.go
@@ -1,8 +1,14 @@
 package rest
 
+import (
+	"fmt"
+	"strings"
+)
+
 type Config struct {
 	Enable     bool   `toml:"enable"`
 	Listen     string `toml:"listen"`
+	BasePath   string `toml:"base_path"`
 	EnableCORS bool   `toml:"enable_cors"`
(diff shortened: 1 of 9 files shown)


www/rest/config.go
```diff
@@ -1,19 +1,42 @@
...
+}
+
+func (c *Config) apiPattern() string {
+	return fmt.Sprintf("%sapi/", c.rootPattern())
+}
+
+func (c *Config) rootPattern() string {
+	path := fmt.Sprintf("/%s/", c.BasePath)
+	path = strings.ReplaceAll(path, "//", "/")
+	path = strings.ReplaceAll(path, "//", "/")
```
ja7ad, line 39: Remove duplicate replace.

ja7ad (changes requested):

www/rest/config.go
b00f: One tests needs it... `///` -> `/`

ja7ad (approved):

PR #1771, opened 2025-05-03
chore(scripts): support docker compose service for snapshot

ja7ad:
## Description

Currently, snapshot supports managing the daemon service via a `systemd` unit. This PR adds support for `systemctl` and Docker Compose services to retrieve snapshots.


b00f:
We can have both daemon and docker in same script

b00f (approved):

themantre (approved):

Issue #1766, opened 2025-04-27
Adding test for `loadOrCreateKey`


themantre:
The `loadOrCreateKey` function in the `network` module needs test coverage:

Test cases needed:
- Valid key should load successfully and Network Key ID should match
- Invalid key should return an error
- Missing key file should return an error

LeonLow97:
Hi @themantre , I am working on this issue 👍

LeonLow97:
Hi @themantre , I have added unit tests and created a pull request for this issue 👍

LeonLow97:
@b00f 

Sure! What is my scope in the development team? What features do I need to work on?

PR #1783, opened 2025-05-10
test(network): added unit test for loadOrCreateKey

LeonLow97:
## Description

> Added unit test for loadOrCreateKey, tested with the following cases:

1. Valid Private Key created
2. Invalid Private Key in existing file
3. Input supplied with directory instead of file path
4. Input supplied with invalid file path

## Related Issue

> Nil

Fixes # (issue)

- Increased Unit Test Coverage from 76.6% to 82.1% in `network_test.go`

### DIFF
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"testing"
 	"time"
 
@@ -512,14 +511,13 @@ func TestLoadOrCreateKey(t *testing.T) {
 	})
 
 	t.Run("Should return error when file contains invalid private key", func(t *testing.T) {
-		tempFile, err := os.CreateTemp("", "invalid_key")
-		assert.NoError(t, err)
+		tempFilePath := util.TempFilePath()
 
 		// Writes an invalid private key to the file, decoding key will fail later
-		err = util.WriteFile(tempFile.Name(), []byte("invalid_data"))
+		err := util.WriteFile(tempFilePath, []byte("invalid_data"))
 		assert.NoError(t, err)
 
-		key, err := loadOrCreateKey(tempFile.Name())
+		key, err := loadOrCreateKey(tempFilePath)
 		assert.Error(t, err)
 		assert.Nil(t, key)
 	})
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"testing"
 	"time"
 
@@ -498,9 +497,8 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
 }
 
 func TestLoadOrCreateKey(t *testing.T) {
-	t.Run("ReturnsSuccessfullyWhenValidKeyCreated", func(t *testing.T) {
-		tempDir := t.TempDir()
-		keyPath := filepath.Join(tempDir, "valid_key")
+	t.Run("Should load same network key after being created", func(t *testing.T) {
+		keyPath := util.TempFilePath()
 
 		// Create new valid key
 		validKey, err := loadOrCreateKey(keyPath)
@@ -510,17 +508,15 @@ func TestLoadOrCreateKey(t *testing.T) {
 		previousValidKey, err := loadOrCreateKey(keyPath)
 		assert.NoError(t, err)
 
-		// Compare public keys of both private keys
-		assert.True(t, validKey.GetPublic().Equals(previousValidKey.GetPublic()))
+		assert.Equal(t, validKey.GetPublic(), previousValidKey.GetPublic())
 	})
 
-	t.Run("ReturnsErrorWhenInvalidKeyCreated", func(t *testing.T) {
+	t.Run("Should return error when file contains invalid private key", func(t *testing.T) {
 		tempFile, err := os.CreateTemp("", "invalid_key")
 		assert.NoError(t, err)
-		defer os.Remove(tempFile.Name())
 
 		// Writes an invalid private key to the file, decoding key will fail later
-		err = os.WriteFile(tempFile.Name(), []byte("invalid_data"), 0644)
+		err = util.WriteFile(tempFile.Name(), []byte("invalid_data"))
 		assert.NoError(t, err)
 
 		key, err := loadOrCreateKey(tempFile.Name())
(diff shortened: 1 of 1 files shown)


LeonLow97:
Resolves #1766

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
 	networkB.Stop()
 	networkP.Stop()
 }
+
+func TestLoadOrCreateKey(t *testing.T) {
+	t.Run("ReturnsSuccessfullyWhenValidKeyCreated", func(t *testing.T) {
+		tempDir := t.TempDir()
+		keyPath := filepath.Join(tempDir, "valid_key")
```
b00f, line 503: Use [utils.TempFilePath](https://github.com/pactus-project/pactus/blob/6b43f89771b5785bbb4507da04a4cd44e5a17f84/util/io.go#L72) instead.

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
 	networkB.Stop()
 	networkP.Stop()
 }
+
+func TestLoadOrCreateKey(t *testing.T) {
+	t.Run("ReturnsSuccessfullyWhenValidKeyCreated", func(t *testing.T) {
```
b00f, line 501: Write proper description for tests. Here for example: "Should create network key if it doesn't exist."

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+
+		// Create new valid key
+		validKey, err := loadOrCreateKey(keyPath)
+		assert.NoError(t, err)
+
+		// Retrieve previously created valid key, the file path exists
+		previousValidKey, err := loadOrCreateKey(keyPath)
+		assert.NoError(t, err)
+
+		// Compare public keys of both private keys
```
b00f, line 513: avoiding over explaining in comments

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+		assert.True(t, validKey.GetPublic().Equals(previousValidKey.GetPublic()))
```
b00f, line 514: user `assert.Equal(t, foo, bar)`

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+	})
+
+	t.Run("ReturnsErrorWhenInvalidKeyCreated", func(t *testing.T) {
+		tempFile, err := os.CreateTemp("", "invalid_key")
+		assert.NoError(t, err)
+		defer os.Remove(tempFile.Name())
```
b00f, line 520: No need to remove Temp file.. Keep test simple

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+
+		// Writes an invalid private key to the file, decoding key will fail later
+		err = os.WriteFile(tempFile.Name(), []byte("invalid_data"), 0644)
```
b00f, line 523: Use `utils.WriteFile`

network/network_test.go
```diff
@@ -494,3 +496,48 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+
+	t.Run("ReturnsErrorWhenGivenDirectory", func(t *testing.T) {
+		tempDir := t.TempDir()
+		key, err := loadOrCreateKey(tempDir)
+		assert.Error(t, err)
+		assert.Nil(t, key)
+	})
+
+	t.Run("ReturnsErrorWithInvalidFilePath", func(t *testing.T) {
+		key, err := loadOrCreateKey("/invalid-file-path")
```
b00f, line 539: Invalid directory in different OS has different pattern.

b00f (changes requested):
The most important test is missing. We want to ensure that the network key is properly loaded with the same network ID

network/network_test.go
LeonLow97: I have changed this to use `util.TempFilePath`

network/network_test.go
LeonLow97: I have changed all unit test titles to follow this format

network/network_test.go
LeonLow97: I have removed this comment

network/network_test.go
LeonLow97: I have changed this to use `assert.Equal` instead of `assert.True`

network/network_test.go
LeonLow97: Good point, I have removed the `defer` statement

network/network_test.go
LeonLow97: I have changed this to use `util.WriteFile`

network/network_test.go
LeonLow97: I have changed this to use `invalidPath := string([]byte{0x00})` which is a null byte is invalid in file names across all OS

LeonLow97:
Hi @b00f , thanks for the informative review, I have addressed the comments.

<br>

> The most important test is missing. We want to ensure that the network key is properly loaded with the same network ID

Regarding this point, I have already included it in the first unit test, i.e., `Should load same network key after being created` where I first create a private key and then load the private key again. I compared both public keys to ensure that they are the same within the same file path. Let me know if I am missing anything here 👍

network/network_test.go
```diff
@@ -494,3 +495,46 @@ func checkConnection(t *testing.T, networkP, networkB *network) {
...
+
+		// Retrieve previously created valid key, the file path exists
+		previousValidKey, err := loadOrCreateKey(keyPath)
+		assert.NoError(t, err)
+
+		assert.Equal(t, validKey.GetPublic(), previousValidKey.GetPublic())
+	})
+
+	t.Run("Should return error when file contains invalid private key", func(t *testing.T) {
+		tempFile, err := os.CreateTemp("", "invalid_key")
```
b00f, line 515: Use utils functions here: `keyPath := util.TempFilePath()`

b00f (changes requested):

b00f:
Thanks @LeonLow97 Only one comment left. Once it resolved we can merge the PR.

LeonLow97:
Awesome @b00f , thanks for the review, I have made the relevant changes 👍

network/network_test.go
LeonLow97: I have changed this to use `util.TempFilePath`

b00f (approved):

LeonLow97:
Unit test seems to be failing on Windows OS that is unrelated to my change, probably an environment configuration issue

```
Running TestNetwork/Reachability_Status
--- FAIL: TestNetwork (14.47s)
    network_test.go:135: TestNetwork seed is 1747036069997689600
    --- FAIL: TestNetwork/Reachability (1.46s)
        network_test.go:256: 
            	Error Trace:	D:/a/pactus/pactus/network/network_test.go:256
            	Error:      	Condition never satisfied
            	Test:       	TestNetwork/Reachability
07:48:04 WRN error on sending msg error="libp2p error: connection failed" _network="<peer.ID 12*3rgoSv> - TestNetwork: {3}" pid=12D3KooWSybER2xyCMyh69tp2kmGidS6Pq2biksMErTgzZ4vQb5F
```

themantre (approved):

PR #1802, opened 2025-06-05
refactor(other): use slices.Contains to simplify code

rustrover:
## Description

There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.


b00f:
@rustrover Thanks

PR #1837, opened 2025-07-23
refactor(util): use slices.Equal to simplify code

minxinyi:
## Description

In the Go 1.21 standard library, a new function has been introduced that enhances code conciseness and readability. It can be find  [here](https://pkg.go.dev/slices@go1.21.1#Equal).

## Related Issue

Fixes # (issue)


b00f (approved):

minxinyi:
Thanks for your review.

PR #1885, opened 2025-09-19
fix(gtk): initial gtk  before using inside startupAssistan funct

phoenixit99:


cmd/gtk/main.go
```diff
@@ -128,15 +134,23 @@ func main() {
...
 	cmd.TrapSignal(func() {
 		cmd.PrintInfoMsgf("Exiting...")
 
-		node.Stop()
-		_ = fileLock.Unlock()
+		if node != nil {
+			node.Stop()
+		}
+		if fileLock != nil {
+			_ = fileLock.Unlock()
```
b00f, line 152: It supposed to unlock automatically

b00f (changes requested):

PR #1904, opened 2025-09-26
feat(network): set limit for network message size

b00f:
## Description

This PR sets limits for network message sizes. Currently, Gossip messages cannot exceed 1 MB (libP2P default), and stream messages are limited to 8 MB.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -158,6 +158,7 @@ linters:
         - ip
         - "no"
         - tt
+        - n
         - i
         - j
         - l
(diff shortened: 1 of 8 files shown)


ja7ad:
Please fix some failure test

ja7ad:
While adding CBOR decode limits is a good defensive step, the current values (`65,536`) may still leave us exposed to memory-bomb payloads since legitimate arrays/maps in our P2P layer are far smaller. We should also consider capping nesting depth and handling indefinite-length items, ensure that violations trigger clear errors and peer penalties, and ideally separate the unrelated `BlockPerSession` change into its own commit with rationale. Tighter, configurable defaults (e.g., a few thousand elements) aligned with actual protocol usage would reduce risk while preserving interoperability.

PR #1975, opened 2025-10-30
chore(version): bumping version to 1.10.2

themantre:
Bumping version to 1.10.2

### DIFF
--- a/version/version.json
+++ b/version/version.json
@@ -2,6 +2,6 @@
     "major": 1,
     "minor": 10,
     "patch": 2,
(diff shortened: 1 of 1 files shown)


version/version.json
```diff
@@ -1,7 +1,7 @@
 {
     "major": 1,
     "minor": 10,
-    "patch": 1,
+    "patch": 2,
     "meta": "",
```
b00f:
```suggestion
    "meta": "beta",
```

Issue #1972, opened 2025-10-30
Update dependencies


b00f:
Project dependencies should be updated. Follow [Update Dependencies](https://github.com/pactus-project/pactus/blob/main/docs/update-dependencies.md) document.

sureshkrishnan-v:
@b00f sure I will start updating the project dependencies

PR #1981, opened 2025-11-13
chore(other): minor improvement for docs

tinyfoolish:
## Description

minor improvement for docs

## Related Issue

Fixes # (issue)


b00f:
@tinyfoolish Thanks.

Do you think you can fix #1972 as well?

tinyfoolish:
@b00f   I have executed the commands according to the documentation, but it seems there are no dependencies to update.


image

image

b00f (approved):

PR #1982, opened 2025-11-19
refactor(version): replace Split in loops with more efficient SplitSeq

efcking:
## Description

strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.


b00f:
Thanks @efcking 

Do you think you can fix this issue as well #1972

efcking:
> Thanks @efcking
> 
> Do you think you can fix this issue as well #1972

Of course. https://github.com/pactus-project/pactus/pull/1984  Please review it.

PR #2138, opened 2026-04-16
feat(state, grpc): calculate the average availability score

b00f:
## Description

Calculate the average availability score.


b00f:
Fixed #2151

PR #2185, opened 2026-05-06
test(sandbox): mocking Sandbox interface

### LINKED ISSUE #2180: Snadbox

b00f:
## Description

Use `gomock` for mocking Sandbox and updating the tests.

## Related Issue

Fixes #2180


b00f (commented):
## Review Summary

This is a significant and well-structured refactoring. Splitting the `Sandbox` interface into `SandboxReader` + `Sandbox` is a clean design, and the `DefaultFactory` pattern for executor injection is a nice approach. However, I found one critical bug and one duplicate in `sandbox_fake.go`.

---

### :red_circle: Critical — `CurrentHeight()` configured twice, making `FakeSandbox.SbxHeight` useless

In `sandbox/sandbox_fake.go`, `NewFakeSandbox` registers `CurrentHeight()` on the mock **twice**:

1. **Line ~95** (correct): Uses `DoAndReturn` to delegate to `fake.SbxHeight` — this allows tests to control the height by setting `fake.SbxHeight`.
2. **Line ~109**: `fake.EXPECT().CurrentHeight().Return(ts.RandHeight()).AnyTimes()` — this call overrides the first, causing `CurrentHeight()` to always return a **static random value**.

**Impact:** Any test that sets `td.sbx.SbxHeight = expiry` (like the delegated unbond test at line ~2533) **will not work**. The `CurrentHeight()` call inside the production code will return the random value snapshotted at construction time, not the test-controlled `SbxHeight`.

**Fix:** Remove line ~109 (`fake.EXPECT().CurrentHeight().Return(ts.RandHeight()).AnyTimes()`). The `DoAndReturn` setup on line ~95 already covers this correctly.

---

### :yellow_circle: Minor — Duplicate `Validator()` EXPECT in `NewFakeSandbox`

The `Validator` mock method is set up twice with identical `DoAndReturn`:
- Lines ~82-86
- Lines ~93-97

The second registration silently shadows the first. Remove one.

---

### :yellow_circle: Minor — Removed test case in `TestUnbondNonDelegatedWithDelegateOwner`

The old test had two sub-tests: the "Should fail, attacker sets DelegateOwner on non-delegated" case, **and** the "Ok, self-unbond with no DelegateOwner still works" case. The latter was removed. Was this intentional?

b00f:
**On `sandbox/sandbox_fake.go` line ~109:**

:red_circle: **Critical bug:** This line overrides the earlier `DoAndReturn` setup for `CurrentHeight()` that was correctly wired to `fake.SbxHeight`. Because gomock uses the **last** registration for the same method call, `CurrentHeight()` will always return a static random value instead of delegating to the settable `fake.SbxHeight` field.

This means the delegated unbond test that sets `td.sbx.SbxHeight = expiry` to advance time will silently break — `CurrentHeight()` will keep returning the value randomized at construction time.

```go
// LINES ~95-100 (correct, but shadowed):
fake.EXPECT().CurrentHeight().DoAndReturn(
    func() types.Height {
        return fake.SbxHeight  // <- This is what tests need
    },
).AnyTimes()

// LINE ~109 (WRONG - overrides the above):
fake.EXPECT().CurrentHeight().Return(ts.RandHeight()).AnyTimes()  // <- Remove this
```

themantre (approved):

PR #2188, opened 2026-05-07
fix(sync): check proposal message for empty proposal

themantre:
## Description

This PR checks if the proposal message is `nil`.


b00f:
fixed #2219

PR #2189, opened 2026-05-07
fix(sync): check vote message for empty vote

themantre:
## Description

This PR checks if the vote message is `nil`.


b00f:
Fixed #2218

PR #2198, opened 2026-05-09
chore(state): log last block version and hash

b00f:
## Description

Add more information about restoring state, including the last block's version and hash.


b00f:
fixed #2217

Issue #944, opened 2024-01-17
GPRC method health check


ja7ad:
## Describe the problem you'd like to have solved

The primary challenge to be addressed involves implementing a reliable and efficient health-check method for a Pactus client node that utilizes gRPC in Golang. 

### Describe the ideal solution

https://pkg.go.dev/google.golang.org/grpc/health

b00f:
@kehiy 
I think health here means to check the gRPC health.

apples-kksk:
/attempt #944

Plan:
- Register the standard gRPC health service on the existing Pactus gRPC server.
- Mark the node-level service and registered Pactus services as SERVING.
- Add matching stream interceptors so health Watch follows the same auth/recovery path as unary RPCs.
- Add bufconn tests for health Check/Watch and BasicAuth coverage.

kachokvolodia:
/attempt #944

Plan:
- Register the standard `grpc.health.v1.Health` service on the existing Pactus gRPC server.
- Mark the default server and registered Pactus services as `SERVING`, including Wallet only when enabled.
- Add stream interceptors so `Health/Watch` follows the same BasicAuth and recovery middleware as unary RPCs.
- Add bufconn/unit tests for `Check`, `Watch`, wallet status, and stream BasicAuth.

PR #2199, opened 2026-05-09
feat(grpc): add health service

apples-kksk:
/claim #944

## Summary
- Register the standard `grpc.health.v1.Health` service on the Pactus gRPC server.
- Report `SERVING` for the overall server and registered Pactus services, including wallet when enabled.
- Add matching stream auth/recovery interceptors so `Watch` follows the same middleware path as unary RPCs.
- Add bufconn coverage for health `Check`, `Watch`, wallet registration, and BasicAuth behavior.

## Tests
- `go test ./www/grpc`
- `go test ./www/...`

## Proof
The new bufconn tests exercise the health service end-to-end without requiring a running node process.

### DIFF
--- a/www/grpc/health_test.go
+++ b/www/grpc/health_test.go
@@ -1,0 +1,126 @@
+package grpc
+
+import (
+	"context"
+	"encoding/base64"
+	"testing"
+
+	pactus "github.com/pactus-project/pactus/www/grpc/gen/go"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


b00f:
@apples-kksk 

apples-kksk:
Thanks for clarifying. That makes sense if the existing ping service is now the preferred way to check gRPC connectivity.

I will leave this closed and avoid opening duplicate health-service PRs for #944. Thanks for taking the time to explain.

PR #2200, opened 2026-05-10
Add standard gRPC health checks

### LINKED ISSUE #944: GPRC method health check

synechism:
## Summary

This adds standard gRPC health checking support to the Pactus gRPC server
using `grpc.health.v1.Health` instead of introducing a custom RPC.

## What Changed

- register the standard gRPC health service during server startup
- mark the root service and all registered Pactus gRPC services as `SERVING`
- include wallet health status when wallet RPC support is enabled
- add tests covering health checks for the default services and wallet service

## Notes

- health checks follow the existing unary interceptor chain, so if gRPC basic auth
  is enabled, health requests are authenticated the same way as other unary RPCs
- no proto changes were needed because this uses the standard gRPC health service

## Testing

```bash
../../tools/go/bin/go test ./www/grpc/...
```

Closes #944

### DIFF
--- a/www/grpc/server.go
+++ b/www/grpc/server.go
@@ -14,6 +14,8 @@ import (
 	pactus "github.com/pactus-project/pactus/www/grpc/gen/go"
 	"github.com/pactus-project/pactus/www/zmq"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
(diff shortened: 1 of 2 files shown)


b00f:
@synechism 

PR #2201, opened 2026-05-10
feat(grpc): add health check service

kachokvolodia:
## Summary
- register the standard `grpc.health.v1.Health` service on the Pactus gRPC server
- report `SERVING` for the default service and registered Pactus services, including Wallet when enabled
- add stream interceptors so `Health/Watch` follows the same BasicAuth and recovery path as unary RPCs
- add bufconn/unit coverage for `Check`, `Watch`, wallet health status, and stream BasicAuth

## Tests
- `go test ./www/grpc`
- `go test ./www/...`

## Proof
The new bufconn tests exercise the health service end-to-end without requiring a running node process. I also ran the broader `www` package suite to cover adjacent HTTP, HTML, basic auth, and ZMQ packages.

/claim #944


b00f:
@kachokvolodia 

PR #2202, opened 2026-05-10
feat(grpc): add health check service

### LINKED ISSUE #944: GPRC method health check

DrakeMikels:
## Summary
- Register the standard gRPC health service for Pactus gRPC servers
- Mark the aggregate service and registered Pactus services as SERVING, including Wallet only when enabled
- Add stream BasicAuth and recovery interceptors so health Watch uses the same middleware coverage as unary calls
- Add tests for health Check, health Watch, wallet status, and BasicAuth-protected health calls

Fixes #944
/claim #944

## Verification
- go test ./www/grpc
- go test ./node

## Demo notes
- `TestHealthCheck` covers the aggregate, Blockchain, Transaction, Network, and Utils health services
- `TestHealthCheckWithWallet` verifies the Wallet health service is only SERVING when Wallet is enabled
- `TestHealthWatchBasicAuth` verifies unauthenticated Watch calls fail and authenticated Watch calls receive SERVING

### DIFF
--- a/www/grpc/middleware.go
+++ b/www/grpc/middleware.go
@@ -15,34 +15,63 @@ func BasicAuth(storedCredential string) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req any, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (
 		any, error,
 	) {
-		user, password, err := htpasswd.ExtractBasicAuthFromContext(ctx)
-		if err != nil {
-			return nil, status.Error(codes.Unauthenticated, "failed to extract basic auth from header")
-		}
-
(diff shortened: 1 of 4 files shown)


DrakeMikels:
Demo video attached for #944. It shows the PR scope and local verification: `go test ./www/grpc` and `go test ./node`.

https://github.com/user-attachments/assets/431a4e97-8e3e-46c0-b155-d7ac651c468d

b00f:
@DrakeMikels 

PR #2205, opened 2026-05-10
feat(grpc): register health check service

marketingapex7:
## Summary

- register the standard gRPC health checking service on the Pactus gRPC server
- report the default service as `SERVING`
- add a bufconn regression test that verifies `Health/Check` returns `SERVING`

## Why

This lets operators and tooling probe the Pactus gRPC server using the standard `grpc.health.v1.Health` API instead of relying on project-specific RPC calls.

## Validation

- `git diff --check`

I could not run `go test` locally because this environment does not have `go`/`gofmt` installed.

/claim #944


b00f:
@marketingapex7 

PR #2231, opened 2026-05-25
feat(grpc): add standard health check service

duongynhi000005-oss:
Adds the standard gRPC health service and a basic server test that verifies SERVING.


b00f:
@duongynhi000005-oss 

PR #2307, opened 2026-06-19
fix(sync): remove gzip compression from blocks response

### LINKED ISSUE #2305: sync: remove compression from blocks response; keep decompression for backward compatibility
## Summary

Compression on `BlocksResponseMessage` yields only ~3% savings (80,752 → 78,228 bytes). Block data is already compact CBOR-encoded binary — cryptographic fields like hashes and signatures are incompressible. The CPU cost outweighs the negligible bandwidth savings.

Removing compression also mitigates the compression bomb concern (related to https://github.com/pactus-project/pactus/pull/2290).

## Changes

1. Remove `bdl.CompressIt()` from `blocksResponseHandler.PrepareBundle` in `sync/handler_blocks_response.go`
2. Optionally remove the `CompressIt()` method from `sync/bundle/bundle.go` (only caller was the above)
3. **Keep** the decompression path in `bundle.go:Decode()` (`BundleFlagCompressed` branch) and `util.DecompressBuffer` — still needed for bundles received from older peers

## Backward Compatibility

Older peers may still send bundles with the `0x0100` compressed flag. The decode path must still handle those until the next milestone.

b00f:
## Description

Removes gzip compression from `BlocksResponseMessage` bundles. Investigation showed only ~3% size reduction (80,752 → 78,228 bytes for 60 blocks). Block data is already compact CBOR-encoded binary — cryptographic fields like hashes and signatures are incompressible. The CPU cost of gzip compress/decompress outweighs the negligible bandwidth savings.

This also mitigates compression bomb concerns.

The decompression path in `Decode()` is preserved for backward compatibility with older peers that still send compressed bundles.

## Related Issue

Fixes #2305

### DIFF
--- a/sync/bundle/bundle.go
+++ b/sync/bundle/bundle.go
@@ -59,10 +59,6 @@ func (b *Bundle) LogString() string {
 	return fmt.Sprintf("%s%s", b.Message.Type(), b.Message.LogString())
 }
 
-func (b *Bundle) CompressIt() {
-	b.Flags = util.SetFlag(b.Flags, BundleFlagCompressed)
-}
-
 type _Bundle struct {
 	Flags           int          `cbor:"1,keyasint"`
 	MessageType     message.Type `cbor:"2,keyasint"`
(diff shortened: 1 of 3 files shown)


b00f:
Superseded by a revised approach. We'll create a new issue and PR with a different strategy.

PR #2323, opened 2026-06-22
feat(gtk): make snapshot url configurable in gui

### LINKED ISSUE #2313: Make snapshot URL configurable in CLI and GTK
## Description

The snapshot URL is currently hardcoded as a constant `DefaultSnapshotURL = "https://snapshot.pactus.org"` in `cmd/importer.go`. While the daemon `import` command already accepts a `--server-addr` flag (defaulting to `DefaultSnapshotURL`), the GTK startup assistant at `cmd/gtk/startup_assistant.go:253` still hardcodes the value directly:

```go
snapshotURL := cmd.DefaultSnapshotURL // TODO: make optional
```

## Proposed Solution

- Ensure the snapshot URL is consistently configurable across both the CLI daemon `import` command and the GTK startup assistant
- The CLI `import` command should use `--server-addr` ~~(already implemented) and/or add a `--snapshot-url` option~~ with default value `https://snapshot.pactus.org`
- The GTK startup assistant should allow users to specify a custom snapshot URL instead of hardcoding it.
- 
## Affected Files

- `cmd/importer.go` - contains the hardcoded `DefaultSnapshotURL` constant
- `cmd/daemon/import.go` - already has `--server-addr` flag, may need review
- `cmd/gtk/startup_assistant.go` - hardcodes the URL (see TODO comment at line 253)

LucasM4r:
## Description

This pull request makes snapshotURL configurable from CLI and GTK
 - Added a new CLI option `--snapshot-url` for the daemon import command, with default value set to the project snapshot default.
 - Kept `--server-addr`  for backward compatibility and marked it as a legacy option.
 - Added precedence logic so `--snapshot-url` is used when both flags are provided.
 - Added a warning when both flags are provided with different values.
 - Removed hardcoded snapshot URL usage from the GTK startup assistant import path.
 - Added GTK startup flag plumbing so snapshot URL can be configured and passed into the assistant flow.
 - Added tests for snapshot URL resolution behavior and precedence cases.

## Related Issue

Closes #2313


cmd/daemon/import.go
```diff
@@ -25,7 +25,9 @@ func buildImportCmd(parentCmd *cobra.Command) {
 
 	workingDirOpt := addWorkingDirOption(importCmd)
 	serverAddrOpt := importCmd.Flags().String("server-addr", cmd.DefaultSnapshotURL,
-		"import server address")
+		"import server address (legacy option; use --snapshot-url instead)")
```
b00f: Why do we need to add legacy support here?

cmd/daemon/import.go
```diff
@@ -134,3 +137,24 @@ func updateProgressBar(bar *progressbar.ProgressBar, fileName string, stats down
 		_ = bar.Finish()
 	}
 }
+
+func resolveSnapshotURL(c *cobra.Command, snapshotURL, serverAddr string) string {
```
b00f: This function just adds complexity to the code, remove it.

cmd/daemon/import_test.go
```diff
@@ -0,0 +1,85 @@
+package main
+
+import (
+	"testing"
+
+	pactuscmd "github.com/pactus-project/pactus/cmd"
+	"github.com/spf13/cobra"
+)
+
+func TestResolveSnapshotURL(t *testing.T) {
```
b00f: Keep code simple then we don't need this test

b00f (changes requested):

b00f:
@LucasM4r Thanks for the PR. Please remove changes from CLI as it is not necessary, add the option flag only for GUI now

cmd/daemon/import.go
LucasM4r: Legacy support was initially added for backward compatibility, but I removed the CLI changes as requested in the review.

cmd/daemon/import.go
LucasM4r: Done

LucasM4r:
Hi @b00f, done. I removed the CLI changes as requested and kept the PR scope GUI-only.

b00f (approved):

b00f:
@LucasM4r Thanks,  check #2326 and #2327 issues as well. It's easy and I am sure you can fix them.

PR #2378, opened 2026-07-10
test(store): mocking store interface

### LINKED ISSUE #2164: Store
Replace the hand-written `MockStore` in `store/mock.go` with a gomock-generated mock from `store/interface.go`.

## Task
- Generate the mock using `mockgen`:
  ```
  mockgen -source=store/interface.go -destination=store/store_mock.go -package=store
  ```
- Ensure it is consistent with the existing code.
- Update all tests that currently use `MockStore` to use the new `MockStore` (gomock-based).

## References
- Completed gomock migrations: #2165 (Consensus), #2166 (State), #2169 (TxPool)
- Parent issue: #2163

b00f:
## Description

Mocking the Store interface

## Related Issue

Fixes #2164


b00f (commented):
## PR Review: test(store): mocking store interface (#2378)

### Overall Assessment

This is a solid refactoring that replaces the hand-written `MockStore` with gomock-generated mocks and "Fake" wrappers. The approach is well-motivated: gomock handles interface boilerplate, while Fake objects provide sensible defaults. This will make maintaining tests much easier as the store interface evolves.

The consistent `Fake` prefix naming convention across fields is a clear improvement over the previous inconsistent naming (`SbxHeight`, `SbxCommittee`, `LastHeight`, etc.).

---

### Critical Issues (Must Fix)

**1. `FakeStore` is missing `PublicKey` and `HasPublicKey` mock expectations**

`store/store_fake.go`

The `PublicKey` and `HasPublicKey` methods from the `Reader` interface are not set up in `NewFakeStore`. These are called by `CommittedBlock.ToBlock()` (in `store/interface.go:39`) and `CommittedTx.ToTx()` (line 68) when restoring stripped public keys from transactions.

If any test calls `ToBlock()` or `ToTx()` on a `CommittedBlock`/`CommittedTx` returned by the FakeStore, the gomock controller will panic with "unexpected call" because no expectation was registered.

**Recommendation**: Add `.AnyTimes()` expectations for `PublicKey` and `HasPublicKey` in `NewFakeStore`, similar to how the old `MockStore` implemented them (iterating over validators and transactions). Alternatively, return `ErrNotFound` gracefully instead of panicking.

---

### Major Issues (Should Fix)

**2. `FakeStore` is missing `IsBanned` and `RecentTransaction` expectations**

`store/store_fake.go`

These are also not set up. Evidence: `state/state_test.go` must add them manually:
```go
fakeStore.EXPECT().IsBanned(gomock.Any()).Return(false).AnyTimes()
fakeStore.EXPECT().RecentTransaction(gomock.Any()).Return(false).AnyTimes()
```

Every consumer of `FakeStore` will need to remember to add these, which is error-prone. Better to include them in `NewFakeStore` with sensible defaults.

**3. `state/state_fake.go` - Write lock used where read lock is sufficient**

`ChainInfo()` (line 241) acquires `fake.lk.Lock()` (a write lock) but only reads data. This unnecessarily serializes concurrent readers. It should use `fake.lk.RLock()` / `fake.lk.RUnlock()` instead.

---

### Minor Issues

**4. `FakeStore` does not deep-clone accounts/validators (behavior change from old MockStore)**

The old `MockStore.Account()` returned `a.Clone()`, protecting internal state. The new `FakeStore` returns the raw pointer. This is a behavior change that could allow tests to accidentally mutate store state through the sandbox. Consider using `Clone()` to preserve the defensive copying behavior.

**5. `sandbox/sandbox_fake.go` - `MakeNewValidator` uses `ts.RandInt32()` for validator number**

```go
fake.EXPECT().MakeNewValidator(gomock.Any()).DoAndReturn(
    func(pub *bls.PublicKey) *validator.Validator {
        return validator.NewValidator(pub, ts.RandInt32())  // random number
    },
).AnyTimes()
```

Using a random number here can make tests non-deterministic. Consider using a sequential counter instead, or accept the number as a parameter.

---

### Suggestions

**6. `state/state_fake.go` - `errFakeInvalid` is still declared but may be unused**

The old `errFakeNotFound` was correctly removed, but `errFakeInvalid` remains declared. Consider checking whether it is still referenced anywhere and removing it if no longer needed.

**7. `consensus/consensus_test.go` and `consensusv2/consensus_test.go` - Shared validator slice across states**

All four states reference the same `vals` slice:
```go
stateX.FakeCommittee.FakeValidators = vals
stateY.FakeCommittee.FakeValidators = vals
...
```

This means mutations to the committee validators of one state affect all states. Consider creating separate copies if each state should have independent committee state.

---

### Summary

| # | Type | Area | Description |
|---|------|------|-------------|
| 1 | Critical | `store/store_fake.go` | Missing `PublicKey`/`HasPublicKey` expectations → will panic on `ToBlock()` |
| 2 | Major | `store/store_fake.go` | Missing `IsBanned`/`RecentTransaction` defaults |
| 3 | Major | `state/state_fake.go` | Write lock used where read lock is sufficient |
| 4 | Minor | `store/store_fake.go` | No deep-cloning (behavior change from old mock) |
| 5 | Minor | `sandbox/sandbox_fake.go` | Random validator number in MakeNewValidator |
| 6 | Suggestion | `state/state_fake.go` | Dead `errFakeInvalid` variable? |
| 7 | Suggestion | `consensus/` tests | Shared mutable validator slice across states |

The overall direction is excellent and this is a maintainability improvement. The critical issue (#1) should be addressed before merging.

b00f (commented):
## Corrected Review: test(store): mocking store interface (#2378)

*(Replaces previous review after closer code inspection)*

### Overall Assessment

This is a solid refactoring replacing the hand-written MockStore with gomock-generated mocks + Fake wrappers. The approach is well-motivated. Naming is now consistent (FakeHeight, FakeStore, FakeCommittee, etc.) vs the old mix of SbxHeight, LastHeight, SbxCommittee, etc.

---

### Retracted Items

After re-examining the code, items #1 and #2 from my previous review are not actual issues:

1. ~~Missing PublicKey/HasPublicKey~~ - The CommittedBlock.store field (unexported) is nil in blocks from FakeStore, but ToBlock() only calls s.store.PublicKey() when a transaction has stripped keys. Test-generated blocks (via GenerateTestBlock) never strip keys, so this code path is never reached. No issue.

2. ~~Missing IsBanned/RecentTransaction~~ - These are intentionally left out so tests can control behavior per-test via direct .EXPECT() calls on the embedded MockStore. This is the correct separation: FakeStore provides defaults for the common path; specific expectations are set in each test.

---

### Remaining Issues

**1. Minor - Dead code: errFakeInvalid is declared but never used**

state/state_fake.go, line 26

```go
var errFakeInvalid = errors.New("invalid")
```

This variable is declared but never referenced anywhere in the codebase. errFakeNotFound was correctly removed, but errFakeInvalid remains as dead code and should be cleaned up.

**2. Minor - ChainInfo() uses write lock where read lock is sufficient**

state/state_fake.go, line 238

ChainInfo() acquires fake.lk.Lock() (write lock) but only reads data (fetches from FakeStore, reads FakeCommittee, etc.). Since other read-only methods like LastBlockHeight() and LastBlockTime() correctly use RLock(), ChainInfo() should too for consistency.

**3. Suggestion - FakeStore no longer deep-clones accounts/validators**

The old MockStore.Account() returned a.Clone(), protecting internal store state. The new FakeStore returns the pointer directly. This is a deliberate simplification, but worth noting as a behavior change for anyone migrating tests - they may now need to clone explicitly if their test mutates accounts retrieved through the sandbox.

**4. Suggestion - FakeSandbox.MakeNewValidator uses random validator number**

sandbox/sandbox_fake.go, line 79

```go
return validator.NewValidator(pub, ts.RandInt32())
```

Using RandInt32() for the validator number makes the validator's identity non-deterministic. If a test relies on validator numbers (e.g., ValidatorByNumber), results will be unpredictable.

---

### Summary

| # | Type | Area | Description |
|---|------|------|-------------|
| 1 | Minor | state/state_fake.go | Dead errFakeInvalid variable |
| 2 | Minor | state/state_fake.go | Write lock where read lock suffices in ChainInfo() |
| 3 | Suggestion | store/store_fake.go | No deep-cloning (behavior change from old mock) |
| 4 | Suggestion | sandbox/sandbox_fake.go | Random validator number in MakeNewValidator |

Good work - the PR is clean and well-structured. No blocking issues.

b00f (commented):
## Updated Review

### Fixed in latest commit

- **Dead code `errFakeInvalid` removed** - The unused variable and its `errors` import were removed from `state/state_fake.go`. ✅
- **Removed commented-out code** - `// committee, _ := td.GenerateTestCommittee(4)` in `consensus/consensus_test.go` and `// assert.Equal(t, ...)` in `state/lastinfo/last_info_test.go` are cleaned up. ✅
- **Removed unused fields** - `fakeCommittee` in `sandbox/sandbox_test.go` and `fakeStore` in `state/state_test.go` test data structs removed. ✅
- **Gomock import style** - All mock files changed from named `gomock "go.uber.org/mock/gomock"` to anonymous `"go.uber.org/mock/gomock"` - cleaner. ✅
- **Extra blank line** - Removed extraneous blank line in `store/store_fake.go` Transaction method. ✅
- **Lint fix** - `http.Post` replaced with `http.NewRequestWithContext` in `www/jsonrpc/server_test.go`. ✅
- **Floating-point comparison** - `assert.Equal` for `LastBlockTime` changed to `assert.InEpsilon` to avoid flaky tests. ✅

### Still open (minor)

- `state/state_fake.go` - `ChainInfo()` uses write lock (`Lock()`) where read lock (`RLock()`) would suffice. Non-blocking.
- `sandbox/sandbox_fake.go` - `MakeNewValidator` uses `ts.RandInt32()` for validator number (non-deterministic).

The PR looks good to me now. No blocking issues.

themantre (approved):

